PR #9, opened 2025-08-14
fix: prettify error printing

sidwebworks:
# Description 📣
This PR adds better error logging and formatting with additional details about the errors.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

---


packages/api/errors.go
```diff
@@ -76,5 +82,47 @@ func NewAPIErrorWithResponse(operation string, res *resty.Response, additionalCo
...
+	if details != nil {
+		apiError.Details = details
+	}
+
 	return apiError
 }
+
+type errorResponse struct {
+	Message string   `json:"message"`
+	Details []string `json:"details"`
```
sheensantoscapadngan, line 94: details isn't necessarily array of string though right?

packages/cmd/tokens.go
```diff
@@ -144,22 +129,24 @@ var tokensCreateCmd = &cobra.Command{
...
+			WorkspaceId: workspaceId,
+			Scopes:      permissions,
+			ExpiresIn:   expireSeconds,
+			Permissions: accessLevels,
+			RandomBytes: hexEncodedRandomBytes,
+
+			// No longer required for creating service tokens:
+			EncryptedKey: "",
+			Iv:           "",
+			Tag:          "",
```
sheensantoscapadngan, line 142: will this break when CLI is the latest version and the Infisical API server is outdated?

DanielHougaard: No this is really old, we removed the encryption key requirement for creating service tokens a long time ago. I highly doubt this will ever be a reported.

This was actually not apart of Sid's PR, but my SRP removal initiative. I think you should fix the PR so the changes are inline with what's in main @sidwebworks

PR #10, opened 2025-08-15
fix(cmd/service-token): create service tokens error

varonix0:
# Description 📣

This PR resolves an issue with creating service tokens. We con't require project private keys to create service token anymore, and in fact we aren't even returning the user private key to the CLI anymore. 

The CLI would attempt to get the users private key (which isn't stored in the CLI), which would then subsequently fail and throw an error during the service token creation step.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/packages/cmd/tokens.go
+++ b/packages/cmd/tokens.go
@@ -4,8 +4,6 @@ Copyright (c) 2023 Infisical Inc.
 package cmd
 
 import (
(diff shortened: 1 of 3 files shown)


sidwebworks (approved):
LGTM

Issue #55, opened 2025-11-08
Add path and tags to .infisical.json config file.


Billcountry:
I find having the default config file quite useful to define how each project operates.

However the current options are quite limiting and still leave the caller having to type `tags` and `path` with every call to the cli. These are still prone to typos when you have to type them often, having them in the config will make the cli more user friendly.

gee-forr:
It would be great if we could also have the API URL in .infisical.json for self-hosted instances, so that we don't have to set an env var.

pandvan:
This would dramatically streamline Infisical adoption to our devs. Once `.infisical.json` is properly configured, they just have to execute `infisical run -- whatever`

PR #56, opened 2025-11-10
Expand config file to support env + tags | Ensure uniformity of loading config accross commands

Billcountry:
# Description 📣

Fixes: #55 

### Expanding config file
- The config file used to support only the `workspaceId` and `env`.
- However, for most cases, as a user I'm also working with `tags` and `path`, and these can be prone to typos if I'm changing them every time, having them in the config helps.
- I understand not every command needs these values,part of the refactor below ensures that these values are only read when the command defines them.

### Code Refactor + Simplification
_I didn't need to do this but I found it necessary to safely introduce the changes above. Here are the motivations around the refactor_

- `GetWorkspaceConfigFromCommandOrFile`
  - I've introduced this file as a simple way of loading the config, it's going to take into account the flags provided in the command, and the contents of the config file, then give you a unified config.
  - This change was necessary because the logic to check if a project ID is provided or if it's in the config was fragmented all over the code but doing the same thing. In some cases even loading the config file more than once in the same function.
  - This unification ensures that every function needing some or the full config that can be in a file can easily get it.

- `Get(String|Boolean|Int|StringSlice)Argument`
  - Being a commandline application, a lot of places have the code
```go
value, err := cmd.Flags().GetString("some-flag")
if err != nil {
  HandleError(err, "A message")
}
```
  - These 4 lines added a lot of broiler plate that can now be simplified to
    - `GetStringArgument(cmd, "some-flag", "Unable to parse flag --some-flag")`

## Type ✨

- [ ] Bug fix
- [x] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️
Added new unit tests testing functionality around `GetWorkspaceConfigFromCommandOrFile`
```sh
go test -v ./packages/util/...
```

Manually run tests against the build:
image

With tags vs without tags in file
image

---

### DIFF
--- a/packages/util/helper.go
+++ b/packages/util/helper.go
@@ -7,18 +7,205 @@ import (
 	"encoding/hex"
 	"fmt"
 	"math/rand"
+	"net"
 	"os"
 	"os/exec"
 	"path"
 	"sort"
 	"strings"
+	"sync"
 	"time"
+
+	"path/filepath"
 
 	"github.com/Infisical/infisical-merge/packages/api"
 	"github.com/Infisical/infisical-merge/packages/models"
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
 )
+
+const (
+	RELAY_CACHE_FILE = "/var/lib/infisical/cached_relay"
+)
+
+func GetRelayName(cmd *cobra.Command, forceRefetch bool, accessToken string) (string, error) {
+	httpClient, err := GetRestyClientWithCustomHeaders()
+	if err != nil {
+		return "", err
+	}
+	httpClient.SetAuthToken(accessToken)
+
+	relayName, err := GetCmdFlagOrEnvWithDefaultValue(cmd, "relay", []string{"INFISICAL_RELAY_NAME"}, "")
+	if err != nil {
+		return "", fmt.Errorf("unable to parse relay flag: %v", err)
+	}
+	if relayName != "" {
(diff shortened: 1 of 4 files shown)


packages/util/folders.go
```diff
@@ -27,14 +27,6 @@ func GetAllFolders(params models.GetAllFoldersParameters) ([]models.SingleFolder
 			loggedInUserDetails = EstablishUserLoginSession()
 		}
 
-		if params.WorkspaceId == "" {
```
Billcountry: Get all folders doesn't need to do these checks.

All commands use `GetWorkspaceConfigFromCommandOrFile` and perform this check in advance.

packages/util/helper.go
```diff
@@ -196,33 +196,6 @@ func RequireServiceToken() {
...
-	workspaceFilePath, err := GetWorkSpaceFromFilePath(projectConfigFilePath)
-	if err != nil {
-		PrintErrorMessageAndExit(fmt.Sprintf("error reading your project config %v", err))
-	}
-
-	if workspaceFilePath.WorkspaceId == "" {
-		PrintErrorMessageAndExit("Your project id is missing in your local config file. Please add it or run again [infisical init]")
-	}
-}
-
```
Billcountry: Checks moved to `GetWorkspaceConfigFromCommandOrFile`

PR #57, opened 2025-11-12
feat: ssh pam

sheensantoscapadngan:
# Description 📣
This PR adds SSH PAM (Privileged Access Management) support to the Infisical CLI, enabling users to establish SSH sessions through the Infisical Gateway with automatic credential injection and session recording.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/packages/cmd/pam.go
+++ b/packages/cmd/pam.go
@@ -75,10 +75,18 @@ var pamDbAccessAccountCmd = &cobra.Command{
 }
 
 var pamSshCmd = &cobra.Command{
-	Use:                   "ssh <account-name-or-id>",
+	Use:                   "ssh",
+	Short:                 "SSH-related PAM commands",
+	Long:                  "SSH-related PAM commands for Infisical",
+	DisableFlagsInUseLine: true,
+	Args:                  cobra.NoArgs,
+}
+
+var pamSshAccessAccountCmd = &cobra.Command{
+	Use:                   "access-account <account-name-or-id>",
 	Short:                 "Start SSH session to PAM account",
 	Long:                  "Start an SSH session to a PAM-managed SSH account. This command automatically launches an SSH client connected through the Infisical Gateway.",
-	Example:               "infisical pam ssh my-server --duration 2h",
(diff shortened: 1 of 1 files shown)


packages/pam/handlers/ssh/proxy.go
```diff
@@ -0,0 +1,406 @@
...
+			Msg("Using password authentication")
+	default:
+		return nil, fmt.Errorf("invalid or unspecified auth method: %s (must be 'public-key' or 'password')", p.config.AuthMethod)
+	}
+
+	// Configure SSH client (proxy acts as client to the target server)
+	clientConfig := &ssh.ClientConfig{
+		User:            p.config.InjectUsername,
+		Auth:            authMethods,
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(), // TODO: Add proper host key verification in production
```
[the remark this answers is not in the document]
sheensantoscapadngan: we'll add this in the future. It's not necessarily urgent because the proxy happens from within a private network

packages/pam/session/logger.go
```diff
@@ -23,8 +23,25 @@ type SessionLogEntry struct {
...
+	TerminalEventInput  TerminalEventType = "input"  // Data from user to server
+	TerminalEventOutput TerminalEventType = "output" // Data from server to user
+)
+
+// TerminalEvent represents a single event in a terminal session
+type TerminalEvent struct {
+	Timestamp   time.Time         `json:"timestamp"`
+	EventType   TerminalEventType `json:"eventType"`
+	Data        []byte            `json:"data"`        // Raw terminal data
+	ElapsedTime float64           `json:"elapsedTime"` // Seconds since session start (for replay)
```
x032205: I get that this could be used for replay, but why not just use timestamp?

x032205: Especially since this is in seconds, which has lower accuracy than MS, making timestamps the better option

x032205 (changes requested):

x032205 (approved):

PR #91, opened 2025-12-16
improvement: add e2e tests poc

fangpenlin:
# Description 📣

ref: https://linear.app/infisical/issue/PLATFRM-125/update-the-current-poc-with-design-doc-feedbacks

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

To run the tests, first, you need to set the path pointing to the [infisical repo's](https://github.com/Infisical/infisical) backend folder as `INFISICAL_BACKEND_DIR` (keep in mind, you need to have feature flags set correctly, we will fix that in the future, but for now it needs to be done manually). For example:

```bash
export INFISICAL_BACKEND_DIR=/Users/fangpenlin/workspace/infisical/backend
```

next, the e2e test by default runs the executable as a subprocess, we need a path t
o the executable of CLI:

```bash
INFISICAL_CLI_EXECUTABLE=/Users/fangpenlin/works
pace/cli/infisical-merge
```

With these, you can then cd into the e2e folder and run the test like this

```bash
cd e2e
go test github.com/infisical/cli/e2e-tests/relay
```

or, you can run the above in just one line:

```bash
INFISICAL_BACKEND_DIR=/Users/fangpenlin/workspace/infisical/backend INFISICAL_CLI_EXECUTABLE=/Users/fangpenlin/works
pace/cli/infisical-merge go test github.com/infisical/cli/e2e-tests/rela
```

To make it possible to trace into the CLI code with a debugger, we also provide `functionCall` as an alt way to run the CLI main function. By doing so, you don't need `INFISICAL_CLI_EXECUTABLE` set. Instead, you need `CLI_E2E_DEFAULT_RUN_METHOD=functionCall`:

```bash
export CLI_E2E_DEFAULT_RUN_METHOD=functionCall
```

Or, you can run it with one line:

```bash
INFISICAL_BACKEND_DIR=/Users/fangpenlin/workspace/infisical/backend CLI_E2E_DEFAULT_RUN_METHOD=functionCall go test github.com/infisical/cli/e2e-tests/relay
```

### DIFF
--- a/e2e/relay/helpers_test.go
+++ b/e2e/relay/helpers_test.go
@@ -232,11 +232,27 @@ func validateExecutable(path string) error {
 	return nil
 }
 
+func getDefaultRunMethod(t *testing.T) RunMethod {
+	envRunMethod := os.Getenv("CLI_E2E_DEFAULT_RUN_METHOD")
+	if envRunMethod == "" {
+		return RunMethodSubprocess
+	}
+
+	// Validate the value
+	runMethod := RunMethod(envRunMethod)
+	if runMethod != RunMethodSubprocess && runMethod != RunMethodFunctionCall {
+		t.Fatalf("CLI_E2E_DEFAULT_RUN_METHOD is set to '%s' but is not a valid run method.\n"+
+			"Valid values are: '%s' or '%s'", envRunMethod, RunMethodSubprocess, RunMethodFunctionCall)
+	}
+
+	return runMethod
+}
+
 func (c *Command) Start(ctx context.Context) {
 	t := c.Test
 	runMethod := c.RunMethod
 	if runMethod == "" {
-		runMethod = RunMethodSubprocess
+		runMethod = getDefaultRunMethod(t)
 	}
 
 	tempDir := t.TempDir()


e2e/relay/relay_test.go
```diff
@@ -0,0 +1,115 @@
...
+		// Ensure the process is still running
+		if !cmd.IsRunning() {
+			slog.Error("Command is not running as expected", "exit_code", cmd.Cmd().ProcessState.ExitCode())
+			cmd.DumpOutput()
+			// Somehow the cmd stops early, let's exit the loop early
+			cmdExit = true
+			return
+		}
+
+		stderr := cmd.Stderr()
```
akhilmhdh: This should be stdout right?

fangpenlin: Nope, stdout is ually for output data or result of the cmd. Usually people use stderr to output the logs or system info. We are using stderr to output the logs, so this is actually correct.

akhilmhdh: Interesting!

akhilmhdh (approved):

PR #93, opened 2025-12-20
improvement: reuse docker compose for e2e tests

fangpenlin:
# Description 📣

ref: https://linear.app/infisical/issue/PLATFRM-134/make-a-smarter-docker-compose-cache-system-to-avoid-recreating-compose

# Tests 🛠️

The way you test this PR is mostly like the #91. The major difference is that now we have a cache system for the docker-compose. If we found a compose (multiple containers with a compose label) already meets the spec, we will not create a new one but reuse it instead. The purpose is to speed up development cycle.

To ensure this system works, you can run the only relay test case we have right now with `TESTCONTAINERS_RYUK_DISABLED=true`. This will stop the [Ryuk container from cleaning up our compose containers](https://golang.testcontainers.org/features/configuration/#customizing-ryuk-the-resource-reaper). After successfully running the relay test case, you can then run again to see how long it takes and if it reuses the existing container.

After that, you can also set `CLI_E2E_DISABLE_COMPOSE_CACHE=1` to disable the cache and you should see it boot the container from the ground up instead of using the cache.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

### DIFF
--- a/e2e/packages/infisical/compose.go
+++ b/e2e/packages/infisical/compose.go
@@ -35,6 +35,69 @@ type BackendOptions struct {
 	Dockerfile string
 }
 
+func (s *Stack) tryReuseExistingContainers(ctx context.Context, uniqueName string) (bool, error) {
+	log.Printf("Trying to reuse existing container: %s", uniqueName)
+	// Try to lookup for existing container with the same name
+	dockerClient, err := testcontainers.NewDockerClientWithOpts(ctx)
+	if err != nil {
+		return false, err
+	}
+	containers, err := dockerClient.ContainerList(ctx, container.ListOptions{
+		All: true,
+		Filters: filters.NewArgs(
+			filters.Arg("label", fmt.Sprintf("%s=%s", api.ProjectLabel, uniqueName)),
+		),
+	})
+	if err != nil {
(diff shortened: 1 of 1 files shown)


e2e/packages/infisical/compose.go
```diff
@@ -22,31 +35,114 @@ type BackendOptions struct {
...
 	if err != nil {
-		return nil, err
+		return err
+	}
+	hashBytes := sha1.Sum(data)
+	hashHex := hex.EncodeToString(hashBytes[:])
+	uniqueName := fmt.Sprintf("infisical-cli-bdd-%s", hashHex)
+
+	// Skip cache lookup if CLI_E2E_DISABLE_COMPOSE_CACHE is set
+	if os.Getenv("CLI_E2E_DISABLE_COMPOSE_CACHE") == "" {
```
akhilmhdh: I think it would be better to create a function and keep it at the top.  The `Up` function seems to have too many logics.

Checking for `CLI_E2E_DISABLE_COMPOSE_CACHE` == “1” and then skipping seems like the right approach.

PR #134, opened 2026-02-23
Improve Update Instructions

pangeran-bottor:
# Description 📣

Fixes incorrect update instructions shown to users based on their install method. Previously, the update hint was hardcoded per OS — for example, Windows users who installed via npm would always see `scoop update infisical` regardless of how they actually installed the CLI.

This PR refactors `GetUpdateInstructions()` to detect the install method at runtime by inspecting the executable path via `os.Executable()` + `filepath.EvalSymlinks()`. Symlink resolution is explicitly handled to cover macOS and Linux where package managers like npm create symlinks in `$PATH` pointing to the real binary inside `node_modules`.

**Before:**
```
Windows → always: scoop update infisical
macOS   → always: brew update && brew upgrade infisical
```

**After:**
```
Windows + npm    → npm update -g @infisical/cli
Windows + scoop  → scoop update infisical
Windows + winget → winget upgrade Infisical.Infisical
macOS + npm      → npm update -g @infisical/cli
macOS + brew     → brew update && brew upgrade infisical
Linux + npm      → npm update -g @infisical/cli
Linux + apt/yum/apk/yay → (unchanged)
```

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️
Added unit tests covering all install method/OS combinations, and multi-OS integration tests that build a test helper binary, place it at real install paths (including via symlink to simulate npm's behavior), and assert the correct instruction is returned.

**Run unit tests:**
```sh
go test ./packages/util/ -run TestGetUpdateInstructions -v
```

**Run integration tests locally (macOS):**
```sh
go build -o update-hint ./test/update-hint

# brew (default)
mkdir -p /tmp/homebrew/bin
cp update-hint /tmp/homebrew/bin/update-hint
/tmp/homebrew/bin/update-hint
# → To update, run: brew update && brew upgrade infisical

# npm via symlink (as npm installs it)
mkdir -p /tmp/node_modules/@infisical/cli/bin /tmp/bin
cp update-hint /tmp/node_modules/@infisical/cli/bin/update-hint
ln -sf /tmp/node_modules/@infisical/cli/bin/update-hint /tmp/bin/update-hint
/tmp/bin/update-hint
# → To update, run: npm update -g @infisical/cli
```

CI runs automatically on `ubuntu-latest`, `macos-latest`, and `windows-latest` via `.github/workflows/test-update-instructions.yml`, covering npm, brew, scoop, winget, and apt detection on their respective platforms.

Sample CI ran successfully: https://github.com/pangeran-bottor/cli/actions/runs/22317820293

---

### DIFF
--- a/packages/util/check-for-update.go
+++ b/packages/util/check-for-update.go
@@ -342,7 +342,7 @@ func getUpdateInstructions(goos string, execPath string) string {
 		if strings.Contains(p, "winget") {
 			return "To update, run: winget upgrade Infisical.Infisical"
 		}
-		return "To update, run: scoop update infisical"
+		return ""
 	case "linux":
 		if isNpm {
 			return "To update, run: npm update -g @infisical/cli"
--- a/packages/util/check-for-update.go
+++ b/packages/util/check-for-update.go
@@ -309,13 +309,44 @@ func getReleasePublishedAt(repoOwner string, repoName string, version string) (t
 }
 
 func GetUpdateInstructions() string {
-	os := runtime.GOOS
-	switch os {
+	execPath, err := os.Executable()
+	if err != nil {
+		execPath = ""
+	}
+	if resolved, err := filepath.EvalSymlinks(execPath); err == nil {
+		execPath = resolved
+	}
+	return getUpdateInstructions(runtime.GOOS, execPath)
+}
+
+func getUpdateInstructions(goos string, execPath string) string {
+	p := strings.ToLower(execPath)
+	isNpm := strings.Contains(p, "node_modules")
+
+	switch goos {
 	case "darwin":
-		return "To update, run: brew update && brew upgrade infisical"
+		if isNpm {
+			return "To update, run: npm update -g @infisical/cli"
+		}
+		if strings.Contains(p, "/homebrew/") || strings.Contains(p, "/cellar/") {
+			return "To update, run: brew update && brew upgrade infisical"
+		}
+		return ""
 	case "windows":
+		if isNpm {
+			return "To update, run: npm update -g @infisical/cli"
+		}
+		if strings.Contains(p, "scoop") {
+			return "To update, run: scoop update infisical"
+		}
+		if strings.Contains(p, "winget") {
+			return "To update, run: winget upgrade Infisical.Infisical"
+		}
 		return "To update, run: scoop update infisical"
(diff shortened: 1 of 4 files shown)


x032205:
@claude review this once

pangeran-bottor:
@x032205 addressed the review feedback in 65da412 and 942eaf2

pangeran-bottor:
hi @x032205 thank you for triggering the re-run. the CI errors seem related with env variable and a flaky test on SSE. but please let me know if there is something that I need to do from my side. thank you!

packages/util/check-for-update.go
```diff
@@ -309,13 +309,44 @@ func getReleasePublishedAt(repoOwner string, repoName string, version string) (t
...
+		if isNpm {
+			return "To update, run: npm update -g @infisical/cli"
+		}
+		if strings.Contains(p, "scoop") {
+			return "To update, run: scoop update infisical"
+		}
+		if strings.Contains(p, "winget") {
+			return "To update, run: winget upgrade Infisical.Infisical"
+		}
 		return "To update, run: scoop update infisical"
```
x032205: I think we should make this fallback be an empty string like the Darwin case

x032205: One problem I see with empty fallback strings is a lack of any information for the user. We could maybe explore linking to the docs rather than showing a command or nothing at all whenever we're unable to detect the installation method. Maybe something like:
```
return "To update, visit https://infisical.com/docs/cli/overview#installation"
```
Needs some experimenting to make it look good though.

x032205: I'll approve either an empty string or a docs fallback! Or if you think we should stick with the current scoop fallback please let me know your reasoning

x032205 (changes requested):

packages/util/check-for-update.go
pangeran-bottor: Agreed, I'll go with the empty string for now. If we get clearer detection paths for the remaining Windows cases (or want to add a docs fallback across all platforms), happy to follow up in a separate PR. Just pushed the change.

x032205 (approved):
Thank you for contributing!

PR #164, opened 2026-03-29
feat: support multiple --path flags in run command

### LINKED ISSUE #900: Specify multiple paths in the CLI
Is it possible to load several secret with `infisical run` ? For instance for all my apps, I want to load secrets from `/global` folder, and specific to the app (`app-A` folder).
Right now I think I'd have to export both secret in a .env and load them from there.
Another solution could be to use secret expension etc to create my `app-A` secret with everything from `global`. But in my use case I don't always want all secrets (it could depends on the environment for example) and I want to keep consistency with commands and secrets accross environment.

I'd love to be able to use --paths with multiple folders.
Maybe my usecase is too specific though?

[Slack Message](https://infisical-users.slack.com/archives/C04BSBMQAQ7/p1692898421654419?thread_ts=1692898421.654419&cid=C04BSBMQAQ7)

vijaygovindaraja:
Fixes Infisical/infisical#900

## Summary

Adds support for specifying multiple `--path` flags in `infisical run` to load and merge secrets from multiple folder paths before injecting them into the subprocess.

**Usage:**
```bash
infisical run --path=/global --path=/app-specific -- npm start
```

This loads secrets from `/global` first, then `/app-specific`. If both paths contain a secret with the same key, the later path's value takes precedence.

**Changes:**
- Changed `--path` flag from `String` to `StringArray` (backwards compatible — single `--path=/` still works as before)
- Added `mergeEnvVars` helper that merges `KEY=VALUE` slices with later entries overriding earlier ones
- Watch mode uses the first path for change detection

## Test plan

- [ ] `infisical run --path=/ -- env` works as before (single path, backwards compatible)
- [ ] `infisical run --path=/global --path=/app -- env` merges secrets from both paths
- [ ] Later paths override earlier paths for duplicate keys
- [ ] `go build ./...` passes

### DIFF
--- a/packages/cmd/run.go
+++ b/packages/cmd/run.go
@@ -157,15 +157,19 @@ var runCmd = &cobra.Command{
 				injectableEnvironment = pathEnvironment
 			} else {
 				// Merge: later paths override earlier paths for duplicate keys
+				prevLen := len(injectableEnvironment.Variables)
 				injectableEnvironment.Variables = mergeEnvVars(injectableEnvironment.Variables, pathEnvironment.Variables)
-				injectableEnvironment.SecretsCount += pathEnvironment.SecretsCount
+				newKeys := len(injectableEnvironment.Variables) - prevLen
+				injectableEnvironment.SecretsCount += newKeys
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


vijaygovindaraja:
@fangpenlin same ask on this one — both PRs are independent so they can be reviewed separately.

PR #175, opened 2026-04-10
fix(cert-manager): await profile slug resolution before certificate issuance

jdoss:
## Summary

- Fix race condition where certificate issuance POST fires before profile slug-to-UUID resolution completes, sending the raw slug as `profileId` and getting a 422
- Merge the slug resolution and certificate monitoring goroutines into a single sequential flow (resolve first, then monitor) in both the `cert-manager agent` and regular `agent` commands
- Reuse existing `lifecycle.max-failure-retries` and `lifecycle.failure-retry-interval` config for initial issuance retries (previously only used for renewal). Default `max-failure-retries: 0` retries indefinitely with exponential backoff capped at 5 minutes

## Bug

The cert-manager agent had a race condition where `POST /api/v1/cert-manager/certificates` fired before the `GET /api/v1/cert-manager/certificate-profiles/slug/{slug}` response returned. This caused the agent to send the profile-name slug string (e.g. `"crdb"`) as the `profileId` field instead of the resolved UUID, resulting in a 422:

```json
{"validation":"uuid","code":"invalid_string","message":"Invalid uuid","path":["profileId"]}
```

### Evidence from server-side logs

The Infisical server received these requests from the same client, 75ms apart:

| Timestamp (ms) | Request ID | Request |
|---|---|---|
| 1775781765199 | req-LiX9p3LQCBiYXr | GET /api/v1/cert-manager/certificate-profiles/slug/crdb |
| 1775781765274 | req-u181LmbJP98hwB | POST /api/v1/cert-manager/certificates (profileId="crdb") |
| 1775781765293 | req-u181LmbJP98hwB | 422 response — Invalid uuid |

The slug lookup hadn't returned yet when the cert issuance request was sent.

## Behavior change

The existing `lifecycle.max-failure-retries` and `lifecycle.failure-retry-interval` config fields now also apply to **initial** certificate issuance, not just renewal. Previously, a failed initial issuance was never retried (the agent was permanently broken until restarted). Now:

- `max-failure-retries: 0` (default) → retry indefinitely with exponential backoff
- `max-failure-retries: N` → retry up to N times then stop
- `failure-retry-interval` → base delay for exponential backoff (default 2s, capped at 5m)

A follow-up docs PR can be opened to document this behavior change.

## Test plan

- [x] `TestResolveCertificateNameReferences` — verifies slug resolution populates `ProfileID` with the UUID
- [x] `TestConcurrentIssuanceBlocksOnSlugResolution` — fires issuance concurrently with a delayed slug resolution, verifies the server rejects the early POST with unresolved ProfileID, proving the ordering guarantee matters
- [x] `TestResolveCertificateNameReferences_MultipleProfiles` — verifies resolution works for multiple certificates with different profiles
- [x] Full project builds cleanly (`go build ./...`)

### DIFF
--- a/packages/cmd/agent.go
+++ b/packages/cmd/agent.go
@@ -2502,32 +2502,32 @@ func (tm *AgentManager) MonitorCertificates(ctx context.Context) {
 		time.Sleep(1 * time.Second)
 	}
 
-	const maxIssuanceAttempts = 3
 	for _, cert := range tm.certificates {
 		cert := cert
 		displayName := tm.getCertificateDisplayName(cert.ID, &cert.Certificate)
-		var lastErr error
-		for attempt := 1; attempt <= maxIssuanceAttempts; attempt++ {
+		maxRetries := cert.Certificate.Lifecycle.MaxFailureRetries
+		retryForever := maxRetries == 0
+
+		baseDelay := 2 * time.Second
+		if d, err := parseDurationWithDays(cert.Certificate.Lifecycle.FailureRetryInterval); err == nil && d > 0 {
+			baseDelay = d
+		}
+
+		const maxBackoff = 5 * time.Minute
+		for attempt := 1; retryForever || attempt <= maxRetries; attempt++ {
 			if err := tm.IssueCertificate(cert.ID, &cert.Certificate); err != nil {
-				lastErr = err
 				log.Error().
 					Str("Certificate", displayName).
 					Int("attempt", attempt).
-					Int("maxAttempts", maxIssuanceAttempts).
 					Msgf("initial certificate issuance failed: %v", err)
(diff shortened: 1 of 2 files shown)


x032205:
@claude review

jdoss:
@x032205 I will fix up the tests.

jdoss:
Ahhh nevermind, it seems PRs from forks don't get these vars.

jdoss:
@x032205 OK I figured out that the initial cert issuance wasn't using the retry logic for renewal, so I wired it up to reuse stuff like

```yaml
  lifecycle:
    renew-before-expiry: "24h"
    status-check-interval: "5m"
    max-failure-retries: 0
    failure-retry-interval: "5s"
```

and I fixed the review notes from Claude bot.

PR #189, opened 2026-04-20
feat: add a reason field before PAM account access

carlosmonastyrski:
# Description 📣

New PAM access reason field introduced on https://github.com/Infisical/infisical/pull/6096.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/packages/api/model.go
+++ b/packages/api/model.go
@@ -847,8 +847,9 @@ type PAMAccessApprovalRequestPayloadRequestData struct {
 }
 
 type PAMAccessApprovalRequest struct {
-	ProjectId   string                                     `json:"projectId"`
-	RequestData PAMAccessApprovalRequestPayloadRequestData `json:"requestData"`
+	ProjectId     string                                     `json:"projectId"`
+	RequestData   PAMAccessApprovalRequestPayloadRequestData `json:"requestData"`
+	Justification string                                     `json:"justification,omitempty"`
 }
 
 type PAMAccessApprovalRequestResponse struct {
--- a/packages/api/model.go
+++ b/packages/api/model.go
@@ -824,6 +824,7 @@ type PAMAccessRequest struct {
 	AccountName  string `json:"accountName,omitempty"`
 	ProjectId    string `json:"projectId,omitempty"`
 	MfaSessionId string `json:"mfaSessionId,omitempty"`
+	Reason       string `json:"reason,omitempty"`
 }
 
 type PAMAccessResponse struct {
--- a/packages/api/model.go
+++ b/packages/api/model.go
@@ -847,9 +847,8 @@ type PAMAccessApprovalRequestPayloadRequestData struct {
 }
 
 type PAMAccessApprovalRequest struct {
-	ProjectId     string                                     `json:"projectId"`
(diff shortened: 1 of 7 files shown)


packages/cmd/pam.go
```diff
@@ -423,6 +455,7 @@ func init() {
 	pamRedisAccessCmd.Flags().String("duration", "1h", "Duration for Redis access session (e.g., '1h', '30m', '2h30m')")
 	pamRedisAccessCmd.Flags().Int("port", 0, "Port for the local Redis proxy server (0 for auto-assign)")
 	pamRedisAccessCmd.Flags().String("project-id", "", "Project ID of the account to access")
+	pamRedisAccessCmd.Flags().String("reason", "", "Reason for accessing the account (stored on the session for audit purposes)")
```
saifsmailbox98: "on the session" can be removed; multiple instances of this present

packages/api/model.go
```diff
@@ -824,6 +824,7 @@ type PAMAccessRequest struct {
 	AccountName  string `json:"accountName,omitempty"`
 	ProjectId    string `json:"projectId,omitempty"`
```
saifsmailbox98, line 825: Ignore comment placement. 

When an access policy is present and the user does not have an active grant then we ask for the reason, and that reason is actually not used. Because the CLI quits after creating the approval request and no session is started.

packages/api/model.go
```diff
@@ -824,6 +824,7 @@ type PAMAccessRequest struct {
...
 	MfaSessionId string `json:"mfaSessionId,omitempty"`
```
saifsmailbox98, line 826: Ignore comment placement. 

Just noticed a discrepancy that in the web access we are not using the reason for justification of the approval request, but in the CLI we are using the reason for the justification of the approval request.

packages/api/model.go
```diff
@@ -824,6 +824,7 @@ type PAMAccessRequest struct {
...
+	Reason       string `json:"reason,omitempty"`
 }
```
saifsmailbox98, line 828: Ignore comment placement. 

ssh exec should be non interactive

saifsmailbox98 (changes requested):

packages/pam/local/ssh-proxy.go
```diff
@@ -41,7 +41,8 @@ func StartSSHLocalProxy(accessToken string, accessParams PAMAccessParams, projec
 
 	pamRequest := accessParams.ToAPIRequest(projectID, durationStr)
 
-	pamResponse, err := CallPAMAccessWithMFA(httpClient, pamRequest)
+	interactive := options.ExecCommand == ""
+	pamResponse, err := CallPAMAccessWithMFA(httpClient, pamRequest, interactive)
```
saifsmailbox98, line 45: Right now, for SSH, we are not asking for the optional reason, both on the CLI and in the web. I think we should only disable this optional prompt for SSH EXEC and not for SSH ACCESS. 

This will make SSH consistent with other resource types.

saifsmailbox98 (changes requested):

saifsmailbox98 (approved):

PR #191, opened 2026-04-22
feat(pam): native RDP client support

bernie-g:
# Description 📣

Adds `infisical pam rdp access` so users can tunnel into PAM-managed Windows targets with their local RDP client. The CLI opens a loopback TCP proxy, the gateway terminates the RDP session with a Rust IronRDP bridge that injects credentials at CredSSP, and all RDP negotiation beyond CredSSP passes through between client and target unchanged.

Pairs with the Windows-unblock PR on the infisical repo: https://github.com/Infisical/infisical/pull/6132

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Manually validated end-to-end with Microsoft Remote Desktop on macOS against a real Windows Server EC2 target: CLI → relay → gateway → Rust bridge → desktop in the client.

```sh
cd packages/pam/handlers/rdp/native && cargo build --release && cd -
go build -tags rdp -o infisical .
infisical pam rdp access --resource <name> --account <name> --project-id <id> --duration 1h --domain <url>
```

---


packages/pam/handlers/rdp/native/src/bridge.rs
```diff
@@ -0,0 +1,316 @@
...
+    }
+}
+
+async fn run_mitm_inner(client_tcp: TcpStream, target: TargetEndpoint) -> Result<()> {
+    // Our tree pulls both ring (direct) and aws-lc-rs (via reqwest); rustls
+    // 0.23 needs an explicit provider when more than one is compiled in.
+    let _ = rustls::crypto::ring::default_provider().install_default();
+
+    let (acceptor_output, connector_output) =
+        tokio::try_join!(run_acceptor_half(client_tcp), run_connector_half(target))?;
```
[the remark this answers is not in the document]
bernie-g: I don't think we need to worry about other processes on the gateway

packages/pam/local/rdp-proxy.go
```diff
@@ -0,0 +1,405 @@
...
+	log.Info().Msgf("RDP connection closed for client: %s", clientConn.RemoteAddr().String())
+}
+
+// writeRDPFile creates a .rdp file pointing at the local loopback
+// listener. Files live under `~/.infisical/rdp/` — matching the CLI's
+// existing convention for per-user state (alongside the login config
+// and update-check cache). Filename includes the session ID so
+// concurrent sessions don't collide; the file is not deleted on exit
+// so users can re-open a session a few times from Finder if they want.
+// Falls back to the OS temp dir if the home directory can't be resolved.
```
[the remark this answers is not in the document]
bernie-g: Fixed the comment

packages/pam/session/uploader.go
```diff
@@ -31,6 +31,10 @@ const (
 	ResourceTypeSSH        = "ssh"
 	ResourceTypeKubernetes = "kubernetes"
 	ResourceTypeMongodb    = "mongodb"
+	// ResourceTypeRDP maps to the backend's PamResource.Windows enum; the
+	// string is "windows" (not "rdp") so the gateway's resource-type tag
```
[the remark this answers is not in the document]
bernie-g: Fixed

packages/pam/pam-proxy.go
```diff
@@ -53,6 +54,7 @@ func GetSupportedResourceTypes() []string {
 		session.ResourceTypeKubernetes,
 		session.ResourceTypeRedis,
 		session.ResourceTypeMongodb,
+		session.ResourceTypeRDP,
```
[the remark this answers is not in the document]
bernie-g: fixed

packages/pam/session/uploader.go
```diff
@@ -31,6 +31,7 @@ const (
 	ResourceTypeSSH        = "ssh"
 	ResourceTypeKubernetes = "kubernetes"
 	ResourceTypeMongodb    = "mongodb"
+	ResourceTypeRDP        = "windows"
```
x032205: Confusing that the const `ResourceTypeRDP` holds the value `"windows"` when every other entry here matches name-to-value (`ResourceTypePostgres = "postgres"`, etc.). Let's either rename to `ResourceTypeWindows` or change the value to `"rdp"` so grepping for either form actually finds the constant.

packages/pam/local/rdp-proxy.go
```diff
@@ -0,0 +1,407 @@
...
+	log.Info().Msgf("RDP proxy server listening on port %d", proxy.port)
+	fmt.Printf("\n")
+	fmt.Printf("**********************************************************************\n")
+	fmt.Printf("                      RDP Proxy Session Started!                      \n")
+	fmt.Printf("----------------------------------------------------------------------\n")
+	fmt.Printf("Resource: %s\n", accessParams.ResourceName)
+	fmt.Printf("Account:  %s\n", accessParams.AccountName)
+	fmt.Printf("\n")
+	fmt.Printf("Connect your RDP client to:\n")
+	util.PrintfStderr("  127.0.0.1:%d\n", proxy.port)
```
x032205, line 117: This block alternates `fmt.Printf` (stdout) and `util.PrintfStderr` (stderr), so section headers like `Connect your RDP client to:` end up on stdout while the actual `127.0.0.1:<port>` and creds go to stderr. Anyone redirecting output (`> out.log`) will see headers with no values underneath. Let's send the whole banner to stderr since it's interactive UI.

packages/pam/pam-proxy.go
```diff
@@ -405,6 +413,24 @@ func HandlePAMProxy(ctx context.Context, conn *tls.Conn, pamConfig *GatewayPAMCo
 		}
 
 		return proxy.HandleConnection(ctx, conn, sessionLogger)
+	case session.ResourceTypeRDP:
+		if credentials.Port < 0 || credentials.Port > 65535 {
```
x032205: Port 0 isn't valid as an outbound TCP destination, so I think this should be `credentials.Port <= 0`. As-is, if the backend sends `Port: 0` (int zero-value when unset) it passes validation and we get a less useful error from the dial later.

x032205 (changes requested):

packages/pam/session/uploader.go
sheensantoscapadngan: this should be `ResourceTypeWindow` since it's a windows server resource right? or would we have WinRM PAM as well in the future which might conflict?

bernie-g: Renamed to ResourceTypeWindows

packages/pam/local/rdp-proxy.go
bernie-g: Fixed

packages/pam/pam-proxy.go
bernie-g: Nice catch, fixed

x032205:
@claude review once

packages/api/model.go
```diff
@@ -890,6 +890,7 @@ type PAMSessionCredentials struct {
 	ServiceAccountToken   string `json:"serviceAccountToken,omitempty"`
 	ServiceAccountName    string `json:"serviceAccountName,omitempty"`
 	Namespace             string `json:"namespace,omitempty"`
+	AccountName           string `json:"accountName,omitempty"`
```
x032205: I don't think we need this. As I pointed out in the other PR, we should use credentials.username instead of the actual account name

packages/pam/handlers/rdp/bridge_cgo_unix.go
```diff
@@ -0,0 +1,195 @@
...
+	if ctrlErr != nil {
+		return -1, ctrlErr
+	}
+	if dupErr != nil {
+		return -1, dupErr
+	}
+	return dup, nil
+}
+
+func (p *RDPProxy) HandleConnection(ctx context.Context, clientConn net.Conn) error {
```
x032205: `HandleConnection` is copy-pasted identically in `bridge_cgo_windows.go`. It doesn't touch any platform-specific APIs, just `StartWithReadWriter` and `Bridge` methods which already live in `bridge_cgo_shared.go`. Let's move it there

.github/workflows/release_build_infisical_cli.yml
```diff
@@ -33,12 +72,14 @@ jobs:
   #     CLI_TESTS_INFISICAL_VAULT_FILE_PASSPHRASE: ${{ secrets.CLI_TESTS_INFISICAL_VAULT_FILE_PASSPHRASE }}
 
   npm-release:
+    if: github.event_name == 'push' || (github.event_name == 'workflow_dispatch' && !inputs.dry_run)
```
x032205: `validate-tag-branch` is now skipped on `workflow_dispatch` (line 22), and it's in `npm-release`'s `needs` without `always()`, so `npm-release` inherits the skip. A manual non-dry-run release would upload goreleaser artifacts but silently skip npm publish. The goreleaser jobs handle this with the `always() && needs.xxx.result` pattern; we should do the same here, or drop `validate-tag-branch` from `npm-release`'s `needs`.

x032205 (changes requested):

.github/workflows/release_build_infisical_cli.yml
bernie-g: Good catch, fixed

lb-vn:
@claude review once

packages/cmd/pam.go
```diff
@@ -415,6 +415,93 @@ var pamRedisAccessCmd = &cobra.Command{
...
+	Long:                  "RDP-related PAM commands for Infisical (Windows Server / Remote Desktop)",
+	DisableFlagsInUseLine: true,
+	Args:                  cobra.NoArgs,
+}
+
+var pamRdpAccessCmd = &cobra.Command{
+	Use:                   "access",
+	Short:                 "Access PAM Windows/RDP accounts",
+	Long:                  "Access a PAM-managed Windows target over RDP. This starts a local loopback proxy that your RDP client connects to; the session tunnels through the Infisical Gateway with credentials injected server-side.",
+	Example:               "infisical pam rdp access --resource windows-prod --account Administrator --duration 1h --project-id <project_uuid>",
```
x032205: nit: `--account Administrator` should be lowercase because we only allow lowercase account names. Should be `--account administrator`

packages/pam/handlers/rdp/native/.gitignore
```diff
@@ -0,0 +1,2 @@
+/target
+/target-docker
```
x032205: Does anything output `/target-docker`? Do we need this or is it unnecessary

.goreleaser-windows.yaml
```diff
@@ -1,13 +1,16 @@
 builds:
```
x032205, line 6: IGNORE PLACEMENT

Claude suggested adding this to the file:
```
release:
  replace_existing_draft: false
  mode: append
```

This is to supposedly add artifacts to the shared draft without overwriting other artifacts.

x032205 (changes requested):
Final comments!

.goreleaser-windows.yaml
bernie-g: Good catch claude. Fixed

packages/pam/handlers/rdp/native/.gitignore
bernie-g: Nope, removed

packages/cmd/pam.go
bernie-g: Fixed

x032205 (approved):

PR #192, opened 2026-04-22
feat(pam): add Oracle DB access support

saifsmailbox98:
# Description 📣

Adds Oracle as a PAM database resource

https://github.com/Infisical/infisical/pull/6134

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


packages/pam/handlers/oracle/proxy_auth.go
```diff
@@ -0,0 +1,891 @@
...
+	// Encrypt REAL password with the real encKey (which equals placeholderEncKey here
+	// because the computation uses session keys + CSK salt only, not the password).
+	newEPassword, err := encryptPassword([]byte(realPassword), realEncKey, true)
+	if err != nil {
+		return nil, fmt.Errorf("encrypt real password: %w", err)
+	}
+
+	// Rebuild the phase-2 payload with substituted AUTH_SESSKEY and AUTH_PASSWORD.
+	rebuilt, err := rebuildPhase2Request(payload, newEClientSessKey, newEPassword)
+	if err != nil {
```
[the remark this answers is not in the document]
saifsmailbox98: fixed. rewriting AUTH_USER in both phase-1 and phase-2 now, matches what postgres/mysql/mssql do

packages/pam/handlers/oracle/proxy_auth.go
```diff
@@ -0,0 +1,891 @@
...
+//	key bytes
+//	val_len (compressed int)
+//	val_len_again (1 byte, same as CLR length)
+//	val bytes
+//	flag (compressed int)
+func replaceKVPValue(payload []byte, key, newValue string) ([]byte, error) {
+	keyBytes := []byte(key)
+	// Search for the key substring; confirm the preceding bytes look like a length prefix.
+	idx := bytes.Index(payload, keyBytes)
+	if idx < 0 {
```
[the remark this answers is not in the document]
saifsmailbox98: fair point, low practical risk (would need literal "AUTH_PASSWORD" inside another field). leaving for now

packages/pam/handlers/oracle/proxy_auth.go
```diff
@@ -0,0 +1,891 @@
...
+	decoded, err := decryptSessionKey(true, placeholderEncKey, p2.EPassword)
+	if err != nil {
+		return nil, fmt.Errorf("decrypt client password: %w", err)
+	}
+	if len(decoded) <= 16 {
+		return nil, fmt.Errorf("decoded password too short")
+	}
+	if string(decoded[16:]) != ProxyPasswordPlaceholder {
+		return nil, fmt.Errorf("password mismatch: got %q", string(decoded[16:]))
+	}
```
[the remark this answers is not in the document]
saifsmailbox98: yep, dropped the %q

packages/pam/handlers/oracle/proxy_auth.go
```diff
@@ -0,0 +1,891 @@
...
+		oldStart := idx + len(keyBytes)
+		oldEnd := valBodyEnd
+		out := make([]byte, 0, len(payload)+len(newValSection))
+		out = append(out, payload[:oldStart]...)
+		out = append(out, newValSection...)
+		out = append(out, payload[oldEnd:]...)
+		return out, nil
+	}
+	return payload, fmt.Errorf("unexpected empty value for %q", key)
+}
```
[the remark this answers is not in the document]
saifsmailbox98: fixed, routed through TTCBuilder.PutClr so long values emit the chunked 0xFE form

packages/pam/handlers/oracle/proxy_auth.go
```diff
@@ -0,0 +1,891 @@
...
+	if !use32Bit {
+		headerLen = 10 // 2 len + 2 ch + 1 type + 1 flag + 2 data = 8 actually. But go-ora and we add dflags.
+		// Actually both use 10 because of the 2-byte data_flags after the 8-byte packet
+		// header. See readDataPayload in o5logon_server.go.
+	}
+	if len(pkt) < headerLen {
+		return nil, fmt.Errorf("packet too short: %d", len(pkt))
+	}
+	return pkt[headerLen:], nil
+}
```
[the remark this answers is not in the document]
saifsmailbox98: gone, collapsed to a single const headerLen = 10

packages/pam/handlers/oracle/proxy_auth.go
```diff
@@ -0,0 +1,891 @@
...
+	// packets first — notably NSPTRS (type 0x0B, "RESEND") which tells the client to
+	// re-transmit its DESCRIPTION as a follow-up packet because it didn't fit inline
+	// in the CONNECT. We forward these intermediates to the client transparently, and
+	// if we see NSPTRS specifically we also read the client's follow-up packet and
+	// forward it back to upstream — otherwise upstream stalls waiting for it.
+	//
+	// A REFUSE / REDIRECT ends the flow with an error.
+	const (
+		PacketTypeResendMarker PacketType = 0x0B // NSPTRS
+	)
```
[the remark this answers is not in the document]
saifsmailbox98: removed, using PacketTypeResend from tns.go

packages/pam/handlers/oracle/proxy.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+	}
+}
+
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
```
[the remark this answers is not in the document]
saifsmailbox98: deleted

packages/pam/handlers/oracle/o5logon.go
```diff
@@ -0,0 +1,136 @@
+// Portions of this file are adapted from github.com/sijms/go-ora/v2,
+// licensed under MIT. Copyright (c) 2020 Samy Sultan.
+// Original: auth_object.go (generateSpeedyKey, getKeyFromUserNameAndPassword,
+// decryptSessionKey, encryptSessionKey, encryptPassword, generatePasswordEncKey) and
+// network/security/general.go (PKCS5Padding).
+// Modifications for server-side use by Infisical: the roles are inverted — the gateway
+// acts as the Oracle server verifying the client's O5Logon using the placeholder password.
```
x032205: Curious if we even need to make these types of comments and have an ATTRIBUTION file for anything under an MIT license?

packages/pam/handlers/oracle/proxy_auth.go
```diff
@@ -0,0 +1,1016 @@
...
+	var acceptVersion uint16
+	if len(acceptRaw) >= 10 {
+		acceptVersion = binary.BigEndian.Uint16(acceptRaw[8:10])
+	}
+	use32Bit := acceptVersion >= 315
+	log.Info().Str("sessionID", p.config.SessionID).Uint16("acceptVersion", acceptVersion).Bool("use32Bit", use32Bit).Msg("Proxy: ACCEPT forwarded")
+
+	// 3. Post-ACCEPT: peek for go-ora's 16-bit-framed connect-data supplement.
+	peekBuf := make([]byte, 256)
+	_ = clientConn.SetReadDeadline(time.Now().Add(3 * time.Second))
```
x032205: This 3-second peek blocks every non-go-ora client (JDBC thin, python-oracledb thin) since they don't send a supplement after ACCEPT. That's a fixed 3s delay on every connection for those clients. We should either shorten this significantly or find a way to detect the client type from the CONNECT packet and skip the peek entirely.

packages/pam/handlers/oracle/query_logger.go
```diff
@@ -0,0 +1,312 @@
...
+	case 900:
+		return "ERROR: ORA-00900: invalid SQL statement"
+	case 942:
+		return "ERROR: ORA-00942: table or view does not exist"
+	case 1017:
+		return "ERROR: ORA-01017: invalid username/password"
+	case 28000:
+		return "ERROR: ORA-28000: the account is locked"
+	}
+	return "ERROR"
```
x032205: This drops the error code for anything not in the switch. We have the code, so let's include it:
```go
return fmt.Sprintf("ERROR: ORA-%05d", code)
```

packages/pam/handlers/oracle/proxy.go
```diff
@@ -0,0 +1,126 @@
...
+
+// OracleProxyConfig mirrors the shape used by other PAM database handlers so the
+// dispatch in pam-proxy.go stays templatized. When EnableTLS is true, the
+// upstream leg uses TLSConfig (built centrally in pam-proxy.go from the
+// resource's sslRejectUnauthorized + sslCertificate fields).
+type OracleProxyConfig struct {
+	TargetAddr     string // "host:port"
+	InjectUsername string
+	InjectPassword string
+	InjectDatabase string
```
x032205, line 16: `InjectDatabase` is set in pam-proxy.go but never read by the oracle handler, we can remove this

packages/pam/handlers/oracle/o5logon.go
x032205: Curious if we even need comment like this (and the ATTRIBUTION) if the source is MIT

packages/pam/handlers/oracle/constants.go
```diff
@@ -0,0 +1,9 @@
+package oracle
+
+// ProxyPasswordPlaceholder is the fixed password string clients must present to the
+// gateway's client-facing O5Logon. Real authentication happens upstream with the real
+// credentials injected by the gateway. The placeholder is not a secret — security is
+// enforced by the mTLS tunnel between CLI, backend and gateway, and by session-scoped
+// client certs. Oracle's O5Logon cannot be bypassed the way MySQL/Postgres auth can,
+// so the gateway and the client must agree on some shared string; this is it.
```
x032205: This comment feels super bloated, I think we can keep it simple and just state that this is a fixed password required to be used by the client.

packages/pam/handlers/oracle/o5logon_server.go
```diff
@@ -0,0 +1,220 @@
...
+	//   user cursor opts (1)
+	//   uol (1)
+	//   sid (4)
+	//   serial num (4)
+	//   rba ts (2)
+	//   rba sqn (4)
+	//   rba blk (4)
+	//   rba byte (4)
+	//   some flags, then CLR of the message
+	// Most fields can be zero; the code is what matters.
```
x032205: I feel like if we really need these comments for each part of the packet, we should put them inline with the p.PutInt() functions below to save space. So instead of having `//   rba byte (4)` on it's own line and `b.PutInt(0, 4, true, true)` down below, we can do:
```
b.PutInt(0, 4, true, true) //   rba byte
```

packages/pam/handlers/oracle/proxy_auth.go
x032205: Try this with a different client like BeeKeeper and let me know if you're able to notice that 3 second delay

packages/pam/handlers/oracle/proxy_auth.go
```diff
@@ -0,0 +1,1016 @@
+package oracle
```
x032205, line 1: Ignore placement but like a good 30% of this 1,000 line file seems to just be comments. While these are supposed to help me understand the code, I find it doing the opposite and just adding more code to review.

The comments are often written with an assumption that I already know the entire flow of the feature, which is not true. This results in me finding most comments useless until after I've already studied the entire feature's code. But at that point there's no need for these comments to even exist, at least not in a verbose state.

We should definitely cut down on any large comment blocks that aren't directly informational. Any info that can be inferred from just understanding the feature/code should not be stated in a comment I think.

Feel free to push back on this in Slack if you disagree. I'm sure different people have different opinions on comments 🤣

x032205 (changes requested):

packages/pam/handlers/oracle/proxy_auth.go
saifsmailbox98: Replaced the timeout with a CONNECT packet inspection. If connect-data-offset + connect-data-length > packet size, the data wasn't inline and a supplement will follow

packages/pam/handlers/oracle/query_logger.go
saifsmailbox98: done

packages/pam/handlers/oracle/proxy.go
saifsmailbox98: That was a miss. We use it now - gateway rewrites SERVICE_NAME in the CONNECT packet to enforce it.

packages/pam/handlers/oracle/o5logon.go
saifsmailbox98: I'm not 100% sure about that, do we have an open source/license/legal expert in the team?

I removed the per-file comments.

Claude:
> trimmed. per-file comments removed, ATTRIBUTION.md carries the MIT license text which is all that's legally required

packages/pam/handlers/oracle/constants.go
saifsmailbox98: removed most of the comments, and made some of them a single line

packages/pam/handlers/oracle/o5logon_server.go
saifsmailbox98: removed the comment, not the most important part of the implementation

packages/pam/handlers/oracle/proxy_auth.go
saifsmailbox98: I added it for AI agents to read, if someone has to later add missing features. Removed most of them.

packages/pam/session/uploader.go
```diff
@@ -31,6 +31,7 @@ const (
 	ResourceTypeSSH        = "ssh"
 	ResourceTypeKubernetes = "kubernetes"
 	ResourceTypeMongodb    = "mongodb"
+	ResourceTypeOracle     = "oracledb"
```
x032205: Maybe we should name it "ResourceTypeOracledb"? To match our existing convention

packages/pam/handlers/oracle/proxy_auth.go
```diff
@@ -0,0 +1,828 @@
...
+			log.Info().Str("sessionID", p.config.SessionID).Int("supplementLen", len(supplement)).Msg("Proxy: forwarding connect-data supplement before handshake")
+			if _, werr := upstreamConn.Write(supplement); werr != nil {
+				return fmt.Errorf("forward connect-data supplement: %w", werr)
+			}
+		}
+	}
+
+	// 3. Read upstream responses until ACCEPT. Handle RESEND (TLS restart).
+	var acceptRaw []byte
+	for attempt := 0; acceptRaw == nil; attempt++ {
```
x032205, line 80: Without any upper bound, is it possible that this loop gets stuck running forever if RESEND keeps being sent?

packages/pam/handlers/oracle/proxy_auth.go
```diff
@@ -0,0 +1,828 @@
...
+
+type ProxyAuthState struct {
+	Salt            []byte
+	Pbkdf2CSKSalt   string
+	Pbkdf2VGenCount int
+	Pbkdf2SDerCount int
+	RealKey         []byte
+	PlaceholderKey  []byte
+	ServerSessKey   []byte
+}
```
x032205, line 554: Claude pointed out that there's no zeroing of the RealKey, PlaceholderKey, and ServerSessKey buffers after auth completes, making them stay in memory for a long time

packages/pam/handlers/oracle/proxy_auth.go
```diff
@@ -0,0 +1,828 @@
...
+
+func translatePhase2Request(payload []byte, state *ProxyAuthState, realPassword string) ([]byte, error) {
+	p2, err := ParseAuthPhaseTwo(payload)
+	if err != nil {
+		return nil, fmt.Errorf("parse client phase 2: %w", err)
+	}
+
+	// server session key
+	// client session key
+	// session1
```
x032205: remove this, i think it was from our call

x032205 (changes requested):

packages/pam/handlers/oracle/proxy_auth.go
saifsmailbox98: removed

packages/pam/handlers/oracle/proxy_auth.go
saifsmailbox98: setting state to nil

packages/pam/handlers/oracle/proxy_auth.go
saifsmailbox98: added an upper bound

packages/pam/session/uploader.go
saifsmailbox98: renamed

x032205 (approved):

PR #195, opened 2026-04-24
feat(pam): add local dev stack for seeding PAM resources

saifsmailbox98:
# Description 📣

Spins up local databases and SSH servers and registers them in Infisical as PAM resources. Useful for testing PAM features and reproducing customer issues.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


dev/pam/resources/postgres/init/seed.sql
```diff
@@ -0,0 +1,1629 @@
...
+
+DO $$
+DECLARE
+    pids UUID[] := ARRAY(SELECT id FROM products);
+    p UUID;
+BEGIN
+    FOR i IN 1..500 LOOP
+        p := pids[1 + floor(random() * array_length(pids, 1))::int];
+        INSERT INTO price_history (product_id, price, changed_at, reason)
+        VALUES (
```
bernie-g, line 433: ignore placement of comment

What are your thoughts on adding some really big tables to these seeds? Would be useful to test edge cases for really long queries/timeouts

dev/pam/resources/redis/users.acl
```diff
@@ -0,0 +1,2 @@
+user default off
```
bernie-g, line 1: ignore comment placement

Can we add some keys to redis? Seems a little weird that it's the only one that doesn't have any seeded data

bernie-g (commented):
ignore placement of comment

dev/pam/resources/redis/users.acl
saifsmailbox98: added some data to redis as well

dev/pam/resources/postgres/init/seed.sql
saifsmailbox98: i actually thought about this, but it would take a lot of space when they run it, and not everyone is working on postgres with huge amounts of data, so it's probably not a good time to add it. we can probably do that once we feel the the need for it, or maybe just add a seed script that can be run manually.

for now, the aim is to keep it simple and lightweight so it's easy for others to pick up, and to see whether it's even worth having in the first place.

dev/pam/resources/mssql/entrypoint.sh
```diff
@@ -0,0 +1,56 @@
...
+
+IF NOT EXISTS (SELECT name FROM sys.database_principals WHERE name = N'$USR')
+    CREATE USER [$USR] FOR LOGIN [$USR];
+
+ALTER ROLE db_owner ADD MEMBER [$USR];
+"
+
+# Run init scripts
+for f in /docker-entrypoint-initdb.d/*.sql; do
+    if [ -f "$f" ]; then
```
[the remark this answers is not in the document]
saifsmailbox98: fixed

dev/pam/Makefile
```diff
@@ -0,0 +1,43 @@
+SHELL := /bin/bash
+
+.PHONY: up down clean setup info
+
+up:
```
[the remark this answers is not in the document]
saifsmailbox98: added the friendly message here as well

dev/pam/setup.sh
```diff
@@ -0,0 +1,308 @@
...
+  if (( ${#rows[@]} > 0 )); then
+    for r in "${rows[@]}"; do echo "$r"; done
+  else
+    echo "(no services currently running — 'make up' brings them up)"
+  fi
+  echo ""
+}
+
+if (( PRINT_INFO_ONLY )); then
+  print_info
```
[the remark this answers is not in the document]
saifsmailbox98: using local01 everywhere

dev/pam/docker-compose.yml
```diff
@@ -0,0 +1,140 @@
...
+      SSH_USER: infisical
+      SSH_MODE: password
+      SSH_PASSWORD: Infisical@123
+    ports:
+      - "127.0.0.1:55422:22"
+
+  ssh-server-key:
+    build:
+      context: ./resources/ssh-server
+    profiles: [ssh-key]
```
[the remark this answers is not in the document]
saifsmailbox98: we can add this later, no need to complicate it for now

dev/pam/docker-compose.yml
```diff
@@ -0,0 +1,140 @@
...
+      MONGO_INITDB_ROOT_USERNAME: infisical
+      MONGO_INITDB_ROOT_PASSWORD: Infisical@123
+      MONGO_INITDB_DATABASE: infisical
+    ports:
+      - "127.0.0.1:55437:27017"
+    volumes:
+      - ./resources/mongodb/init:/docker-entrypoint-initdb.d
+      - pam_mongodata:/data/db
+    healthcheck:
+      test: ["CMD", "mongosh", "--eval", "db.adminCommand('ping')"]
```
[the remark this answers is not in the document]
saifsmailbox98: we can add this later, no need to complicate it for now

dev/pam/resources/postgres/init/seed.sql
bernie-g: Hm ok fair enough

dev/pam/setup.sh
```diff
@@ -0,0 +1,308 @@
...
+
+AUTH_HEADER="Authorization: Bearer ${INFISICAL_TOKEN}"
+PREFIX="${RESOURCE_PREFIX}-$(date +%F)"
+
+# -----------------------------------------------------------------------------
+# Gateway ID — required by the Infisical PAM create-resource endpoints
+# -----------------------------------------------------------------------------
+if [[ -z "${INFISICAL_GATEWAY_ID:-}" ]]; then
+  echo "INFISICAL_GATEWAY_ID is not set. Fetching gateways from ${INFISICAL_DOMAIN} ..."
+  gw_resp=$(curl -sS -H "$AUTH_HEADER" "${INFISICAL_DOMAIN}/api/v1/gateways/")
```
bernie-g: Why does this hit v1?

saifsmailbox98: this should have been v2, but removed the gateway id fetching completely

making it more user friendly could add more points of failure when using and harder to debug

dev/pam/docker-compose.yml
```diff
@@ -0,0 +1,149 @@
+services:
+  postgres:
+    image: postgres:17-alpine
+    profiles: [postgres]
+    environment:
+      POSTGRES_DB: infisical
+      POSTGRES_USER: infisical
+      POSTGRES_PASSWORD: Infisical@123
```
bernie-g, line 8: Thoughts on standardizing all the passwords to just `infisical`? That's the password for the local infisical db and it would be nice to keep it consistent + extra simple to remember

saifsmailbox98: I started with `infisical` initially, but mssql had a different requirement: https://learn.microsoft.com/en-us/sql/relational-databases/security/password-policy?view=sql-server-ver17#password-complexity

the policy can be turned off though, but doing that might lead to unintended behavior for someone testing something related to auth, etc.

could have used `infisical` for others though, but that might create more confusion, so changed all of them to `Infisical@123`, another advantage of this is that if you're looking for places where password/credentials are defined you can do so with a search for `Infisical@123` - `infisical` might match a lot more places.

let me know what you think

bernie-g: Ah ok in that case I think Infisical@123 is fine for consistency. Probably not worth the hassle

bernie-g (approved):

PR #198, opened 2026-04-28
feat: support fetching imported certificates in cert-agent

carlosmonastyrski:
# Description 📣

Adds support in the certificate agent for fetching an existing certificate by ID and writing the cert, chain, and private key in the desired destinations instead of always issuing a new certificate from a profile.
A new `imported-certificate-id` config field selects this mode and skips the issuance flow entirely, periodic status checks still run so revocation and expiry are detected, but renewal is left to whatever issued the cert externally. When the imported certificate has no private key (e.g. ACME-issued), the agent logs a warning if a `private-key` path is configured and otherwise stays quiet.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/packages/api/api.go
+++ b/packages/api/api.go
@@ -62,6 +62,7 @@ const (
 	operationCallInstanceRelayHeartBeat            = "CallInstanceRelayHeartBeat"
 	operationCallIssueCertificate                  = "CallIssueCertificate"
 	operationCallRetrieveCertificate               = "CallRetrieveCertificate"
+	operationCallGetCertificateBundle              = "CallGetCertificateBundle"
 	operationCallRenewCertificate                  = "CallRenewCertificate"
 	operationCallGetCertificateRequest             = "CallGetCertificateRequest"
 )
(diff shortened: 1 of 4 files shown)


bernie-g:
@claude review once

packages/api/api.go
```diff
@@ -1143,6 +1144,25 @@ func CallRetrieveCertificate(httpClient *resty.Client, certificateId string) (*R
 	return &resBody, nil
 }
 
+func CallGetCertificateBundle(httpClient *resty.Client, certificateId string) (*CertificateBundleResponse, error) {
```
bernie-g, line 1147: ignore placement of comment

as far as I can tell there's no way to get the ID of a certificate besides copying it from the URL which seems like subpar UX. Maybe in a different PR we could add a Copy ID button to the frontend

bernie-g (approved):
Tested this and seems to work. Could be worth putting this in the docs as well.

packages/api/api.go
carlosmonastyrski: Yes, I'll raise a separate PR with the doc changes and the UI copy ID button 👍

PR #200, opened 2026-04-29
feat(gateway): direct-to-bucket session recording uploads with reconciliation

x032205:
# Description 📣

Direct-to-bucket session recording uploads with reconciliation

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/packages/api/api.go
+++ b/packages/api/api.go
@@ -57,6 +57,8 @@ const (
 	operationCallUploadPamSessionLog               = "CallUploadPamSessionLog"
 	operationCallPAMSessionTermination             = "CallPAMSessionTermination"
 	operationCallUploadPamSessionEventBatch        = "CallUploadPamSessionEventBatch"
+	operationCallPAMSessionChunkPresignedPut       = "CallPAMSessionChunkPresignedPut"
+	operationCallPAMSessionChunkMetadata           = "CallPAMSessionChunkMetadata"
 	operationCallGetMFASessionStatus               = "CallGetMFASessionStatus"
 	operationCallOrgRelayHeartBeat                 = "CallOrgRelayHeartBeat"
 	operationCallInstanceRelayHeartBeat            = "CallInstanceRelayHeartBeat"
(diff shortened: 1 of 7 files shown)


packages/api/api.go
```diff
@@ -1051,6 +1053,48 @@ func CallUploadPamSessionLogs(httpClient *resty.Client, sessionId string, reques
 	return nil
 }
 
+func CallPAMSessionChunkPresignedPut(
```
bernie-g, line 1057: ignore placement of comment

We should add a warning to the settings page or ideally a hard-check for CORS to make sure to avoid hitting a CORS error during log reads

packages/api/api.go
```diff
@@ -1051,6 +1053,48 @@ func CallUploadPamSessionLogs(httpClient *resty.Client, sessionId string, reques
...
+		SetHeader("X-Gateway-Upload-Token", uploadToken).
+		SetBody(req).
+		SetResult(&resp).
+		Post(fmt.Sprintf("%v/v1/pam/sessions/%s/chunks/presigned-put", config.INFISICAL_URL, sessionId))
+	if err != nil {
+		return ChunkPresignedPutResponse{}, NewGenericRequestError(operationCallPAMSessionChunkPresignedPut, err)
+	}
+	if httpResp.IsError() {
+		return ChunkPresignedPutResponse{}, NewAPIErrorWithResponse(operationCallPAMSessionChunkPresignedPut, httpResp, nil)
+	}
```
bernie-g, line 1075: ignore placement of comment

I don't think reconcilliation works. I tested by having the first upload fail, then restarting the gateway and it doesn't attempt to reupload the logs - the chunks end up stranded on disk

packages/pam/session/uploader.go
```diff
@@ -304,8 +314,17 @@ func readFromOffset(filename, encryptionKey string, offset int64) ([]byte, int64
 			return nil, newOffset, fmt.Errorf("failed to decrypt record at offset %d: %w", newOffset, err)
 		}
 
+		entrySize := len(decryptedData)
+		if len(entries) > 0 {
+			entrySize++ // comma separator
+		}
+		if maxPayloadBytes > 0 && len(entries) > 0 && runningSize+entrySize > maxPayloadBytes {
```
bernie-g, line 321: ignore placement of comment

Currently in the gateway if it tries to upload a chunk that's too big it just errors out. This PR eases that problem a lot by making the chunk limit much higher for S3, but the issue still exists in theory. I think it would be smart to change the gateway to automatically split a chunk up to the maximum size allowed if it's bigger, for both pathways

packages/pam/session/uploader.go
```diff
@@ -488,7 +521,47 @@ func (su *SessionUploader) flushSession(sessionID, encryptionKey string) error {
...
+			}
+
+			currentOffset = newOffset
+			startElapsedMs = endElapsedMs
+		}
+
+		if currentOffset != state.fileOffset {
+			state.lastEndElapsedMs = endElapsedMs
+			state.fileOffset = currentOffset
+			if err := writePersistedOffset(state.filename, currentOffset, endElapsedMs); err != nil {
```
bernie-g, line 546: I think this should be inside the loop, otherwise there could be desync between the offset written to disk and the one stored in memory (`currentOffset`). Otherwise if there's a crash during upload they'll get desynced and will upload the same events again on reconcilliation

bernie-g:
@claude review once

packages/pam/session/uploader.go
```diff
@@ -264,10 +272,27 @@ func deletePersistedOffset(filename string) {
 	_ = os.Remove(offsetFilePath(filename))
 }
 
+func splitPayload(data []byte, maxSize int) [][]byte {
```
bernie-g: This splits by raw bytes which could split the an event in half and result with invalid JSON which would break the playback viewer

packages/pam/session/uploader.go
```diff
@@ -264,10 +272,27 @@ func deletePersistedOffset(filename string) {
...
+	if len(data) <= maxSize {
+		return [][]byte{data}
+	}
+	var parts [][]byte
+	for len(data) > 0 {
+		end := maxSize
+		if end > len(data) {
+			end = len(data)
```
bernie-g: Claude seems to think there's still a race condition where data loss is possible during a crash:

### Summary

`LoadRecordingSecretsFromDisk` at `uploader.go:454-455` is gated by `HasPendingChunks`. If the gateway crashes with an empty queue, the persisted upload token never loads on restart, even though it's on disk.

  The gateway then calls `/credentials`, which returns a valid `sessionKey` but empty `uploadToken` for Active sessions (`pam-account-service.ts:1305`). Any unflushed events in `.enc` get encrypted, queued, and fail to upload (no token to mint
  presigned PUT URLs).

  ### Repro
  1. Start session, run commands, wait for flush ticker → queue empty
  2. Run a few more commands (within the 10s window)
  3. `kill -9` gateway, restart
  4. Step 2's events strand as a `.chunk` file, never reach S3

  ### Impact
  Silent loss of up to ~10s of events per crash. Operator sees a "complete" recording missing data.

  ### Fix
  Drop the `HasPendingChunks` gate — always load the persisted token if the file exists.

packages/pam/session/uploader.go
x032205: You're right, our frontend expects each chunk to be independently parsable. I'm just going to remove the chunk splitting logic and instead warn when chunks are over 64MB (but still let them upload). The hard limit we set on the backend is 256MB if I'm not mistaken, which should never be reached in practice. But if it somehow does, that warning will be there to point out that the size is large (above 64MB).

I don't think it's worth maintaining chunk splitting logic since we need to account for log formats to preserve integrity.

bernie-g: Sounds good

bernie-g (approved):

PR #229, opened 2026-05-11
feat: expose PROJECT_ID via environment variable

MattCrossingham:
Fixes #213

This PR allows users to pass the project ID via the `INFISICAL_PROJECT_ID` or `PROJECT_ID` environment variables across all CLI commands.

Previously, users using CI/CD pipelines had to explicitly pass `--projectId=...` as a flag, even if they had `PROJECT_ID` exported in their environment (unlike `INFISICAL_TOKEN` which was automatically picked up). This update uses `GetCmdFlagOrEnvWithDefaultValue` to cleanly fall back to the environment variables if the flag is omitted.

### DIFF
--- a/packages/cmd/dynamic_secrets.go
+++ b/packages/cmd/dynamic_secrets.go
@@ -43,7 +43,7 @@ func getDynamicSecretList(cmd *cobra.Command, args []string) {
 		util.HandleError(err, "Unable to parse flag")
 	}
 
-	projectId, err := cmd.Flags().GetString("projectId")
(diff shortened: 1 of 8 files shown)


MattCrossingham:
Thanks @chatgpt-codex-connector! You are absolutely right. I've removed the generic `PROJECT_ID` fallback to prevent those collisions. It now strictly looks for `INFISICAL_PROJECT_ID` to match the `INFISICAL_TOKEN` pattern.

scott-ray-wilson:
Thank you for your contribution!

scott-ray-wilson (approved):
LGTM

PR #233, opened 2026-05-13
feat: cert agent applications rework

carlosmonastyrski:
# Description 📣

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/certificate-agent-config.yaml
+++ b/certificate-agent-config.yaml
@@ -1,4 +1,4 @@
-version: v2 # v2 issues from a profile attached to a PKI Application. Use v1 for the legacy project-based flow.
+version: v2 # v2 issues from a profile attached to an Application. Use v1 for the legacy project-based flow.
 
 infisical:
   address: "https://app.infisical.com/"
--- a/certificate-agent-config.yaml
+++ b/certificate-agent-config.yaml
@@ -1,4 +1,4 @@
-version: v1
+version: v2 # v2 issues from a profile attached to a PKI Application. Use v1 for the legacy project-based flow.
 
 infisical:
   address: "https://app.infisical.com/"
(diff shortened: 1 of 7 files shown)


certificate-agent-config.yaml
```diff
@@ -1,4 +1,4 @@
-version: v1
+version: v2 # v2 issues from a profile attached to a PKI Application. Use v1 for the legacy project-based flow.
```
saifsmailbox98: We don't need to use the word `PKI`

packages/cmd/agent.go
```diff
@@ -2032,43 +2040,120 @@ func validateCertificateLifecycleConfig(certificates *[]AgentCertificateConfig)
...
+			matched = &profiles[j]
+			break
 		}
+	}
+	if matched == nil {
+		return fmt.Errorf("profile '%s' is not attached to application '%s'", cert.ProfileName, cert.ApplicationName)
+	}
 
-		cert.ProfileID = profile.ID
+	if matched.APIConfigID == nil || *matched.APIConfigID == "" {
```
saifsmailbox98: can we not rely on the backend throwing an error? we'll have to otherwise keep these two always in sync

saifsmailbox98 (changes requested):

packages/cmd/agent.go
carlosmonastyrski: Makes sense 👍

certificate-agent-config.yaml
carlosmonastyrski: 👍

packages/api/model.go
```diff
@@ -138,19 +138,37 @@ type GetProjectByIdResponse struct {
 
 type GetProjectBySlugResponse Project
 
+type PkiApplication struct {
+	ID        string `json:"id"`
+	Name      string `json:"name"`
+	ProjectID string `json:"projectId,omitempty"`
```
saifsmailbox98: not being used

saifsmailbox98 (approved):

PR #237, opened 2026-05-18
feat(relay): add resource auth enrollment for relays

saifsmailbox98:
# Description 📣

Add --enroll-method flag to relay start command for token and AWS auth. Companion to Infisical/infisical#6515.

Backend PR: https://github.com/Infisical/infisical/pull/6515

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/packages/cmd/relay.go
+++ b/packages/cmd/relay.go
@@ -45,6 +45,10 @@ var relayStartCmd = &cobra.Command{
 		if enrollMethod == "" {
 			enrollMethod = os.Getenv("INFISICAL_RELAY_ENROLL_METHOD")
 		}
+		if enrollMethod != "" && enrollMethod != relay.EnrollMethodToken && enrollMethod != relay.EnrollMethodAws {
+			util.HandleError(fmt.Errorf("invalid --enroll-method %q: supported values are %q and %q",
+				enrollMethod, relay.EnrollMethodToken, relay.EnrollMethodAws))
+		}
 
 		host, _ := util.GetCmdFlagOrEnv(cmd, "host", []string{gatewayv2.RELAY_HOST_ENV_NAME})
 		if host == "" && enrollMethod == "" {
@@ -126,6 +130,8 @@ var relayStartCmd = &cobra.Command{
 				domain, _ := cmd.Flags().GetString("domain")
 				if domain != "" {
 					config.INFISICAL_URL = util.AppendAPIEndpoint(domain)
+				} else if storedDomain, _ := relay.LoadStoredDomain(relayName); storedDomain != "" {
+					config.INFISICAL_URL = util.AppendAPIEndpoint(storedDomain)
 				}
 
 				httpClient, err := util.GetRestyClientWithCustomHeaders()
--- a/packages/cmd/relay.go
+++ b/packages/cmd/relay.go
@@ -12,6 +12,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/Infisical/infisical-merge/packages/api"
+	"github.com/Infisical/infisical-merge/packages/config"
 	gatewayv2 "github.com/Infisical/infisical-merge/packages/gateway-v2"
 	"github.com/Infisical/infisical-merge/packages/relay"
 	"github.com/Infisical/infisical-merge/packages/util"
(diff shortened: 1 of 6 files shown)


packages/cmd/relay.go
```diff
@@ -39,29 +41,166 @@ var relayStartCmd = &cobra.Command{
...
+			storedEnrollToken, _ := relay.LoadStoredEnrollmentToken(relayName)
+			alreadyEnrolled := storedEnrollToken != "" && storedEnrollToken == enrollToken
+
+			if alreadyEnrolled {
+				log.Info().Msg("Enrollment token matches stored token. Skipping enrollment.")
+			} else {
+				domain, _ := cmd.Flags().GetString("domain")
+				if domain != "" {
+					config.INFISICAL_URL = util.AppendAPIEndpoint(domain)
+				}
```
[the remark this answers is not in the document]
saifsmailbox98: Added the LoadStoredDomain fallback in the token enrollment path, same as the AWS path already does

packages/cmd/relay.go
```diff
@@ -39,29 +41,166 @@ var relayStartCmd = &cobra.Command{
...
-		if err != nil || host == "" {
-			util.HandleError(err, fmt.Sprintf("unable to get host flag or %s env", gatewayv2.RELAY_HOST_ENV_NAME))
+		enrollMethod, _ := cmd.Flags().GetString("enroll-method")
+		if enrollMethod == "" {
+			enrollMethod = os.Getenv("INFISICAL_RELAY_ENROLL_METHOD")
+		}
+
+		host, _ := util.GetCmdFlagOrEnv(cmd, "host", []string{gatewayv2.RELAY_HOST_ENV_NAME})
+		if host == "" && enrollMethod == "" {
+			util.HandleError(fmt.Errorf("please provide host flag"), fmt.Sprintf("unable to get host flag or %s env", gatewayv2.RELAY_HOST_ENV_NAME))
```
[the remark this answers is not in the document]
saifsmailbox98: Invalid values are now rejected

saifsmailbox98:
@claude review once

packages/relay/enroll.go
```diff
@@ -0,0 +1,142 @@
...
+			trimmed := strings.TrimSpace(line)
+			if trimmed == "" || strings.HasPrefix(trimmed, prefix) {
+				continue
+			}
+			existingLines = append(existingLines, line)
+		}
+	}
+
+	existingLines = append(existingLines, fmt.Sprintf("%s=%s", key, value))
+	content := strings.Join(existingLines, "\n") + "\n"
```
carlosmonastyrski, line 86: A SIGKILL mid-write would leave the conf empty or truncated, and the next start would fail with “no stored access token found.”

packages/relay/enroll.go
```diff
@@ -0,0 +1,142 @@
...
+
+	INFISICAL_RELAY_ACCESS_TOKEN_KEY    = "INFISICAL_RELAY_ACCESS_TOKEN"
+	INFISICAL_RELAY_DOMAIN_KEY          = "INFISICAL_RELAY_DOMAIN"
+	INFISICAL_RELAY_ENROLLMENT_TOKEN_KEY = "INFISICAL_RELAY_ENROLLMENT_TOKEN"
+	INFISICAL_RELAY_ID_KEY              = "INFISICAL_RELAY_ID"
+)
+
+func relayConfPath(name string) (string, error) {
+	if os.Geteuid() == 0 {
+		return filepath.Join("/etc/infisical/relays", name+".conf"), nil
```
carlosmonastyrski, line 23: should we add a client-side regex check on name matching the backend’s slug patter? (names like "../foo" may cause some troubles here)

carlosmonastyrski (changes requested):

packages/relay/enroll.go
saifsmailbox98: carried it over from initial gateway enrollment code. what else can we do here, a write temp + rename could probably leave sensitive data in a temp file on crash. I think it should be fine to have it like this. User can generate a new token and start over.

packages/relay/enroll.go
saifsmailbox98: the user running the cli command can also cat those files if they want to, so there should ideally be no privilege escalation with this. Also, the name goes to the backend first which only accepts slugs, so no writes would happen to any other location.

carlosmonastyrski (approved):

PR #238, opened 2026-05-18
feat: add principals flag to cli for ssh dynamic secrets

saifsmailbox98:
# Description 📣

Adds a `--principals` flag to infisical dynamic-secrets lease create so users can specify which principals to include in SSH dynamic secret leases. Also adds principals support to the agent dynamicSecret template function as an optional 6th argument.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/packages/cmd/agent.go
+++ b/packages/cmd/agent.go
@@ -49,7 +49,7 @@ const DEFAULT_INFISICAL_CLOUD_URL = "https://app.infisical.com"
 
 const CACHE_TYPE_KUBERNETES = "kubernetes"
 
-const DYNAMIC_SECRET_LEASE_TEMPLATE = "dynamic-secret-lease-%s-%s-%s-%s-%s"
+const DYNAMIC_SECRET_LEASE_TEMPLATE = "dynamic-secret-lease-%s-%s-%s-%s-%s-%s"
 
 // duration to reduce from expiry of dynamic leases so that it gets triggered before expiry
 const DYNAMIC_SECRET_PRUNE_EXPIRE_BUFFER = -15
@@ -273,6 +273,7 @@ type DynamicSecretLeaseWithTTL struct {
 	Data              map[string]interface{}
 	TemplateIDs       []int
 	RequestedLeaseTTL string
+	Principals        string
 }
 
 func (c *CacheManager) WriteToCache(key string, value interface{}, ttl *time.Duration) error {
@@ -399,6 +400,7 @@ func (d *DynamicSecretLeaseManager) WriteLeaseToCache(lease *DynamicSecretLeaseW
 		lease.SecretPath,
 		lease.Slug,
 		requestedLeaseTTL,
+		lease.Principals,
 	)
 
 	ttl := time.Until(lease.ExpireAt)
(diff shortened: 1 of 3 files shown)


packages/cmd/agent.go
```diff
@@ -49,7 +49,7 @@ const DEFAULT_INFISICAL_CLOUD_URL = "https://app.infisical.com"
 
 const CACHE_TYPE_KUBERNETES = "kubernetes"
 
-const DYNAMIC_SECRET_LEASE_TEMPLATE = "dynamic-secret-lease-%s-%s-%s-%s-%s"
+const DYNAMIC_SECRET_LEASE_TEMPLATE = "dynamic-secret-lease-%s-%s-%s-%s-%s-%s"
```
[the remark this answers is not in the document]
saifsmailbox98: looks like we've changed this cache key format before without versioning, old entries just expire via badger TTL

packages/cmd/agent.go
```diff
@@ -399,6 +400,7 @@ func (d *DynamicSecretLeaseManager) WriteLeaseToCache(lease *DynamicSecretLeaseW
 		lease.SecretPath,
 		lease.Slug,
 		requestedLeaseTTL,
+		lease.Principals,
 	)
 
 	ttl := time.Until(lease.ExpireAt)
```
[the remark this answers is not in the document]
saifsmailbox98: Even if this happens, the only effect is an extra lease on the server, both work fine and expire normally

packages/cmd/agent.go
```diff
@@ -503,7 +505,8 @@ func (d *DynamicSecretLeaseManager) DeleteUnusedLeasesFromCache() error {
 				s.Environment == cachedLease.Environment &&
 				s.SecretPath == cachedLease.SecretPath &&
 				s.Slug == cachedLease.Slug &&
-				s.RequestedLeaseTTL == cachedLease.RequestedLeaseTTL
+				s.RequestedLeaseTTL == cachedLease.RequestedLeaseTTL &&
+				s.Principals == cachedLease.Principals
 
 			if match {
 				log.Debug().Msgf("[cache]: found matching active lease: [project=%s], [env=%s], [path=%s], [slug=%s]",
```
adilsitos: don't we also need to add the principals here? isn't it good for debug?

packages/cmd/agent.go
```diff
@@ -942,22 +946,25 @@ func dynamicSecretTemplateFunction(accessToken string, dynamicSecretManager *Dyn
...
-		dynamicSecretData := dynamicSecretManager.GetLeaseUnsafe(accessToken, projectSlug, envSlug, secretPath, slug, templateId, ttl)
+		dynamicSecretData := dynamicSecretManager.GetLeaseUnsafe(accessToken, projectSlug, envSlug, secretPath, slug, templateId, ttl, principals)
 
 		// if a lease is found (either in memory or in cache), we register the template and return the data
 		if dynamicSecretData != nil {
-			dynamicSecretManager.RegisterTemplateUnsafe(projectSlug, envSlug, secretPath, slug, templateId, ttl)
+			dynamicSecretManager.RegisterTemplateUnsafe(projectSlug, envSlug, secretPath, slug, templateId, ttl, principals)
 
-			etagData := fmt.Sprintf("%s-%s-%s-%s-%s", projectSlug, envSlug, secretPath, slug, ttl)
+			etagData := fmt.Sprintf("%s-%s-%s-%s-%s-%s", projectSlug, envSlug, secretPath, slug, ttl, principals)
```
adilsitos, line 970: I think for the cache we won't have a problem if the principals are different, but can't this be a problem on the etag as well? if they change the order of the principals, this would cause a difference an we would write on disk the change (which in case wouldn't really be a change, since only the order was changed) 

Not sure if this is a problem, I just wanted to raise this so we can discuss.

packages/cmd/agent.go
```diff
@@ -980,19 +987,34 @@ func dynamicSecretTemplateFunction(accessToken string, dynamicSecretManager *Dyn
...
+			for _, p := range strings.Split(principals, ",") {
+				trimmed := strings.TrimSpace(p)
+				if trimmed != "" {
+					parsedPrincipals = append(parsedPrincipals, trimmed)
+				}
+			}
+			if len(parsedPrincipals) > 0 {
+				leaseConfig["principals"] = parsedPrincipals
+			}
+		}
```
adilsitos, line 999: maybe this could be a helper? if we have something similar in another command we might also need something similar.

packages/cmd/agent.go
saifsmailbox98: I think it should be fine because if a new lease was issued, then we should probably put that in the file rather than having the old one. The old one would expire on its own by the TTL. 

And this should only happen if somebody is manually updating the template and the principal order, so it looks like they deliberately want that.

packages/cmd/agent.go
saifsmailbox98: yup, added here as well

packages/cmd/agent.go
saifsmailbox98: extracted a helper `ParsePrincipals`

adilsitos (approved):

PR #239, opened 2026-05-19
feat(cli): add smoke tests across multiple distros

x032205:
# Description 📣

Add CLI smoke tests across multiple distros

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


.github/workflows/run-cli-smoke-tests.yml
```diff
@@ -0,0 +1,118 @@
...
+          path: artifacts
+
+      - name: Run smoke tests on ${{ matrix.distro }}
+        run: |
+          chmod +x artifacts/smoke-tests/smoke.sh
+          docker run --rm \
+            -v "${{ github.workspace }}/artifacts/smoke-tests/dist:/dist" \
+            -v "${{ github.workspace }}/artifacts/smoke-tests/smoke.sh:/smoke.sh" \
+            ${{ matrix.image }} \
+            sh -c "${{ matrix.install }} /dist/${{ matrix.pkg }} && /smoke.sh"
```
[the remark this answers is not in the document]
x032205: @claude that's outdated

lb-vn:
@claude review this PR once

lb-vn:
@claude review this once

.github/workflows/run-cli-smoke-tests.yml
```diff
@@ -0,0 +1,169 @@
...
+
+  smoke-test:
+    name: ${{ matrix.distro }}
+    needs: build
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - distro: ubuntu-22.04
```
bernie-g, line 121: The old smoke tests also ran on `ubuntu:20.04, rockylinux:8, amazonlinux:2, and distroless`. The original issue we ran into with the dynamically linked libc only happened on ubuntu 20 and below, so we should at least add that. The other ones probably aren't as important but they run in parallel so I don't see the harm in adding them

smoke-tests/smoke.sh
```diff
@@ -0,0 +1,82 @@
...
+BINARY_PATH=$(command -v infisical)
+if command -v ldd >/dev/null 2>&1; then
+  ldd_output=$(ldd "$BINARY_PATH" 2>&1 || true)
+  if echo "$ldd_output" | grep -qi "not a dynamic executable\|statically linked\|not a valid dynamic program"; then
+    pass "static binary (no dynamic dependencies)"
+  elif echo "$ldd_output" | grep -qi "not found"; then
+    fail "missing shared libraries"
+    echo "$ldd_output"
+  else
+    pass "no missing shared libraries"
```
bernie-g: I think this should be fail, no? We don't know what shared libraries are available on what distros so we should just be disallowing any dynamically linked libraries at all, not relying on guess and check if they're available or not.

bernie-g (changes requested):

bernie-g (approved):

PR #240, opened 2026-05-20
feat: add login status command

Thiago-AS:
# Description 📣

- DOC update: https://github.com/Infisical/infisical/pull/6570

```
Reports whether the CLI is authenticated to Infisical and, when available, the organization the active session is scoped to.

Usage:
  infisical login status

Examples:
infisical login status

Flags:
  -h, --help           help for status
      --json           Output the login status as JSON
      --token string   Inspect this machine identity access token instead of the active session or environment variables

Global Flags:
      --domain string      Point the CLI to your Infisical instance (e.g., https://eu.infisical.com for EU Cloud, or https://your-instance.com for self-hosted). Can also set via INFISICAL_API_URL environment variable. Required for non-US Cloud users. (default "https://app.infisical.com/api")
  -l, --log-level string   log level (trace, debug, info, warn, error, fatal)
      --silent             Disable output of tip/info messages. Useful when running in scripts or CI/CD pipelines.
      --telemetry          Infisical collects non-sensitive telemetry data to enhance features and improve user experience. Participation is voluntary (default true)
```

Screenshot 2026-05-21 at 12 04 46
Screenshot 2026-05-21 at 12 05 00

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/packages/cmd/login_status.go
+++ b/packages/cmd/login_status.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/Infisical/infisical-merge/packages/api"
 	"github.com/Infisical/infisical-merge/packages/config"
 	"github.com/Infisical/infisical-merge/packages/util"
 	"github.com/fatih/color"
@@ -18,9 +17,12 @@ import (
 )
 
 const (
-	statusAuthenticated   = "authenticated"
-	statusExpired         = "expired"
-	statusUnauthenticated = "unauthenticated"
+	statusAuthenticated = "authenticated"
+	statusExpired       = "expired"
+
+	principalKindUser            = "user"
+	principalKindMachineIdentity = "machine-identity"
+	principalKindServiceToken    = "service-token"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


victorvhs017 (changes requested):
The user login works great!

But if I try after logging in with a machine identity, I get:

Image

We should support machine identities, too

packages/cmd/login_status.go
```diff
@@ -39,110 +51,271 @@ var loginStatusCmd = &cobra.Command{
...
+func buildMachineIdentityContext(token, source, domain string) loginStatusContext {
+	// Service tokens (`st.<id>.<key>` format) are opaque — no JWT to decode.
+	if strings.HasPrefix(token, "st.") {
+		return loginStatusContext{
+			kind:        principalKindServiceToken,
+			domain:      domain,
+			rawToken:    token,
+			tokenSource: source,
+		}
+	}
```
victorvhs017: The function name here is a bit misleading; machine identities are not the same as service tokens.

packages/cmd/login_status.go
```diff
@@ -39,110 +51,271 @@ var loginStatusCmd = &cobra.Command{
...
+}
 
-	if loggedInUserDetails.LoginExpired {
-		os.Exit(1)
+// detectMachineIdentityEnvToken returns the machine-identity / service-token
+// credential exported in the environment, mirroring the precedence used by
+// util.GetInfisicalToken. The legacy `TOKEN` gateway variable is intentionally
+// omitted here because its name collides with too many unrelated tools.
+func detectMachineIdentityEnvToken() (token, source string, ok bool) {
```
victorvhs017: same here

packages/cmd/login_status.go
```diff
@@ -155,61 +328,171 @@ func writeLoginStatusJSON(out loginStatusJSONOutput) error {
...
 	}
 
-	util.PrintfStdout("  %s Logged in as %s\n",
-		greenStyle.Sprint("✓"), boldStyle.Sprint(ctx.loggedInUser.UserCredentials.Email))
+	if status != statusAuthenticated {
+		switch ctx.kind {
+		case principalKindUser:
+			util.PrintlnStdout("  - Run `infisical login` to re-authenticate.")
+		case principalKindMachineIdentity:
+			util.PrintlnStdout("  - Run `infisical login` to re-authenticate and re-export your token.")
```
victorvhs017: This may be only a domain issue, like this example, where I expected this to work with the same domain as my logged-in user, but I needed to add the --domain flag for it:

Image

We could update the message to also tell the user to check the domain and update with --domain if needed

packages/cmd/login_status.go
victorvhs017, line 1: If I'm using a JWT token from a user saved in the INFISICAL_TOKEN, it's saying that I'm logged in as a machine identity:

Image

victorvhs017 (changes requested):

packages/cmd/login_status.go
Thiago-AS: Like the public doc state, if all operations using machine/service token will either use the `INFISICAL_API_URL` or the `--domain` flag. I kept the same behavior from other commands. This for me is confusing in the entire CLI not only here.

victorvhs017: The behavior is fine. I think we just need to update the message on the last bullet point to make it clear that they may not need to export the token, just pass the domain. (if that's the issue)

packages/cmd/login_status.go
Thiago-AS: So the doc explicitly consents that the env var and --token are for machine identity usage
image
That's why I made that assumption, will look a way to deterministically tell the types apart.

victorvhs017: This is from the user authentication section in the docs:

image

Thiago-AS: The CLI is too broad haha, already found a way to tell them apart, working in the fix

victorvhs017 (approved):

PR #244, opened 2026-05-27
feat(gateway): match systemd service name to gateway name

bernie-g:
## Summary
The systemd service name now matches the gateway name directly (e.g. `my-gateway.service` instead of the hardcoded `infisical-gateway.service`), with config at `/etc/infisical/gateways/<name>.conf`. Legacy installs are detected and updated in place with a warning to migrate. The `systemd uninstall` command now requires a gateway name argument.

Companion docs PR: https://github.com/Infisical/infisical/pull/6624

## Test plan
- [x] All install/uninstall edge cases tested in Docker with systemd

### DIFF
--- a/packages/cmd/gateway.go
+++ b/packages/cmd/gateway.go
@@ -708,7 +708,7 @@ var gatewaySystemdUninstallCmd = &cobra.Command{
 	Long:                  "Uninstall and remove systemd service for the gateway. Must be run with sudo on Linux.",
 	Example:               "sudo infisical gateway systemd uninstall my-gateway",
 	DisableFlagsInUseLine: true,
-	Args:                  cobra.ExactArgs(1),
+	Args:                  cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		if runtime.GOOS != "linux" {
 			util.HandleError(fmt.Errorf("systemd service uninstallation is only supported on Linux"))
@@ -716,6 +716,13 @@ var gatewaySystemdUninstallCmd = &cobra.Command{
 
 		if os.Geteuid() != 0 {
 			util.HandleError(fmt.Errorf("systemd service uninstallation requires root/sudo privileges"))
+		}
+
+		if len(args) == 0 {
+			if err := gatewayv2.UninstallLegacyGatewaySystemdService(); err != nil {
+				util.HandleError(err, "Failed to uninstall systemd service")
+			}
+			return
 		}
 
 		if err := gatewayv2.UninstallGatewaySystemdService(args[0]); err != nil {
--- a/packages/cmd/gateway.go
+++ b/packages/cmd/gateway.go
@@ -571,7 +571,7 @@ var gatewaySystemdCmd = &cobra.Command{
 	Short: "Manage systemd service for Infisical gateway",
 	Long:  "Manage systemd service for Infisical gateway. Use 'systemd install' to install and enable the service.",
 	Example: `sudo infisical gateway systemd install my-gateway --token=<token> --domain=<domain>
-  sudo infisical gateway systemd uninstall`,
+  sudo infisical gateway systemd uninstall my-gateway`,
 	DisableFlagsInUseLine: true,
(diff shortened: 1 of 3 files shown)


packages/gateway-v2/systemd.go
```diff
@@ -11,20 +11,82 @@ import (
...
+	gatewaysConfigDir  = "/etc/infisical/gateways"
+)
+
+func serviceFilePath(name string) string {
+	return fmt.Sprintf("/etc/systemd/system/%s.service", name)
+}
+
+func gatewayConfigPath(name string) string {
+	return filepath.Join(gatewaysConfigDir, name+".conf")
+}
```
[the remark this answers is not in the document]
bernie-g: If you're running this command you already have root access

packages/gateway-v2/systemd.go
```diff
@@ -11,20 +11,82 @@ import (
...
+
+	if legacy.exists && legacy.gatewayName == name {
+		return installResult{
+			serviceName: legacyServiceName,
+			configPath:  legacyConfigPath,
+			isLegacy:    true,
+		}
+	}
+
+	if legacy.exists {
```
saifsmailbox98, line 68: if the user selects `infisical-gateway` as the name and there is a legacy service for an existing gateway, it will indicate that this service `will be installed alongside it`, but will overwrite the legacy service file.

packages/gateway-v2/systemd.go
```diff
@@ -11,20 +11,82 @@ import (
...
+	if _, err := os.Stat(legacyServicePath); os.IsNotExist(err) {
+		return legacyInfo{}
+	}
+
+	name, _ := readKeyFromConfFile(legacyConfigPath, GATEWAY_NAME_ENV_NAME)
+	return legacyInfo{exists: true, gatewayName: name}
+}
+
+func logLegacyWarning(svcName string) {
+	log.Warn().Msgf("Using legacy service name '%s'. To migrate to the new naming format, run: sudo infisical gateway systemd uninstall %s && sudo infisical gateway systemd install %s <original-flags>", legacyServiceName, svcName, svcName)
```
saifsmailbox98: `<original-flags>` is not very helpful. we can ask them to get a new command/enrollment token from Infisical dashboard

packages/cmd/gateway.go
```diff
@@ -675,38 +681,44 @@ var gatewaySystemdInstallCmd = &cobra.Command{
...
 var gatewaySystemdUninstallCmd = &cobra.Command{
-	Use:                   "uninstall",
+	Use:                   "uninstall [name]",
 	Short:                 "Uninstall and remove systemd service for the gateway (requires sudo)",
 	Long:                  "Uninstall and remove systemd service for the gateway. Must be run with sudo on Linux.",
-	Example:               "sudo infisical gateway systemd uninstall",
+	Example:               "sudo infisical gateway systemd uninstall my-gateway",
 	DisableFlagsInUseLine: true,
-	Args:                  cobra.NoArgs,
+	Args:                  cobra.ExactArgs(1),
```
saifsmailbox98: `infisical gateway systemd uninstall` (wihtout any args) will break on CLI upgrade, not sure if someone might be doing it through scripts.

can we use `use cobra.MaximumNArgs(1)` instead? so uninstall legacy if present, else show error that name is needed?

saifsmailbox98 (changes requested):

packages/cmd/gateway.go
bernie-g: I doubt anyone is running this in a script but yea we can do that

bernie-g: Done

packages/gateway-v2/systemd.go
bernie-g: Sure, done

packages/gateway-v2/systemd.go
bernie-g: Good catch, fixed!

saifsmailbox98 (approved):

PR #247, opened 2026-05-28
feat: allow `--tag` in `infisical secrets set` 

mathnogueira:
# Description 📣

This allows the CLI to create/update a secret via `infisical secrets set` and set its tags with `--tag`.

It allows multiple `--tag` to be set. Following the same pattern of other tools, if the secret already exists and has tags, providing `--tag` will override the existing tags. If no `--tag` is provided, the existing tags will be preserved.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

* Create a new secret with one tag
* Create a new secret with multiple tags
* Create multiple secrets with multiple tags
* Updating a secret that has no tags
* Updating a secret that has existing tags and see that they got overwritten
* Don't pass `--tag` to an existing secret with tags and see that its tags got preserved

```sh
```

---

### DIFF
--- a/packages/api/api.go
+++ b/packages/api/api.go
@@ -1364,3 +1364,42 @@ func CallGetCertificateRequest(httpClient *resty.Client, certificateRequestId st
 
 	return &resBody, nil
 }
+
+func GetTagBySlug(httpClient *resty.Client, projectId string, tagSlug string) (SecretTag, error) {
+	var resBody GetTagBySlugResponse
+	response, err := httpClient.
+		R().
+		SetResult(&resBody).
+		SetHeader("User-Agent", USER_AGENT).
+		Get(fmt.Sprintf("%v/v1/projects/%s/tags/slug/%s", config.INFISICAL_URL, url.PathEscape(projectId), url.PathEscape(tagSlug)))
+
+	if err != nil {
+		return SecretTag{}, NewGenericRequestError("GetTagBySlug", err)
+	}
+
+	if response.IsError() {
+		return SecretTag{}, NewAPIErrorWithResponse("GetTagBySlug", response, nil)
+	}
+
+	return resBody.Tag, nil
+}
+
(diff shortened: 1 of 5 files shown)


packages/api/api.go
```diff
@@ -1364,3 +1364,42 @@ func CallGetCertificateRequest(httpClient *resty.Client, certificateRequestId st
...
+
+	if err != nil {
+		return SecretTag{}, NewGenericRequestError("CreateTag", err)
+	}
+
+	if response.IsError() {
+		return SecretTag{}, NewAPIErrorWithResponse("CreateTag", response, nil)
+	}
+
+	return resBody.Tag, nil
```
adilsitos, line 1404: nit: I know that it is the pattern in the other requests, but maybe `%v` can be too broad? The type of the URL is string, so I don't see why to use `%v`. Maybe we can change in the new ones, but up to you.

packages/api/model.go
```diff
@@ -229,11 +229,12 @@ type Project struct {
...
-	SecretValue   string `json:"secretValue,omitempty"`
-	Type          string `json:"type,omitempty"`
-	SecretComment string `json:"secretComment,omitempty"`
-	ID            string `json:"id,omitempty"`
+	SecretKey     string   `json:"secretKey,omitempty"`
+	SecretValue   string   `json:"secretValue,omitempty"`
+	Type          string   `json:"type,omitempty"`
+	SecretComment string   `json:"secretComment,omitempty"`
+	ID            string   `json:"id,omitempty"`
+	TagIDs        []string `json:"tagIds,omitempty"`
```
adilsitos, line 237: I believe you also need to open a pull request to update the docs on the main repo.

adilsitos (dismissed):
LGTM. just need to add the docs on the main repo if it doesn't exist.

adilsitos (approved):

PR #252, opened 2026-06-02
fix(gateway): --domain flag not working for non-enroll method

x032205:
# Description 📣

Old enrollment method gateway commands such as:
```
infisical relay start --name=test --token=... --domain=http://localhost:8080 --host=1.1.1.1
```

Did not work.

The --domain flag never gets used, and when removed the logged in domain never gets used either. It's ALWAYS app.infisical.com.

This change makes it prioritize domains in the following order:
- `--domain` flag
- StoredDomain (only if using the new enrollment method)
- Logged in domain

This fix makes domains work for both the old and new enrollment methods.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/packages/cmd/gateway.go
+++ b/packages/cmd/gateway.go
@@ -230,6 +230,14 @@ var gatewayStartCmd = &cobra.Command{
 			util.HandleError(errors.New("gateway name is required (provide as positional argument)"))
 		}
 
+		if flagDomain, _ := cmd.Flags().GetString("domain"); flagDomain != "" {
+			config.INFISICAL_URL = util.AppendAPIEndpoint(flagDomain)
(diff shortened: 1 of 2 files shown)


lb-vn:
Removed that check because some users may enroll once and then remove the enroll method flag. That way we'd need to check if the flag either exists OR they already have an enrollment token. That check isn't needed, because if the enrollment token exists then the storedDomain also exists 🤣 

Therefore just checking if a storedDomain exists is the closest we can get to optimal logic

lb-vn:
@greptile re-review this pr

saifsmailbox98 (approved):

PR #258, opened 2026-06-12
feat(export): add --to-files to write per-path .env files

### LINKED ISSUE #1068: Export multiple `.env` files following logical paths
### Feature description
I'd like an option in the `export` command that would create multiple `.env` files following the `path` option (if the option is not provided, it will create a `.env` for all the paths).

#### Example

If I have these folders for grouping my secrets:
- apps/cli
- apps/api
- apps/web
- packages/eslint-config

And I do: `infisical export --path="apps/*" --to-files` it would create these files:
- apps/cli/.env
- apps/api/.env
- apps/web/.env

### Why would it be useful?
It would be super useful for handling secrets in monorepos. Right now we would need to add a custom script in every apps/packages in the monorepo to reflect the logical path into the physical path. For example within my `apps/cli` folder I have to do `infisical export --path="apps/cli > .env` which is not practical when you have a lot of apps and packages.

lollinng:
## Description

Adds a `--to-files` flag to `infisical export`. Closes Infisical/infisical#1068.

Today `export` writes a single stream (stdout, or one file via `--output-file`). In a monorepo you have to run a separate `infisical export --path="apps/cli" > apps/cli/.env` for every app and package to reflect the logical secret-folder layout into the physical filesystem. `--to-files` does that in one command, mirroring each logical path into the filesystem:

- `infisical export --path="apps/*" --to-files` writes `apps/cli/.env`, `apps/api/.env`, `apps/web/.env` (one file per immediate subfolder of `apps`)
- omitting `--path` exports every folder in the tree, one file per path
- a concrete `--path="/apps/cli"` writes the single `apps/cli/.env`

The output filename follows `--format` (`.env`, `secrets.json`, `secrets.yaml`, ...). It reuses the existing formatters, `util.GetAllFolders`, and the per-path secret fetch, and is mutually exclusive with `--output-file`.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How has this been tested?

The path-resolution logic is split into pure helpers (`expandToFilePaths`, `mapPathToFile`) so it can be unit-tested without a live backend. `packages/cmd/export_to_files_test.go` (testify table tests) covers: glob to immediate children only, root-glob to top-level (non-recursive), omitted/root to a full recursive walk, concrete path to itself, leading/trailing-slash normalization, and the per-format filename mapping (12 cases).

- `go build ./...` passes
- `go test -vet=off -run 'TestExpandToFilePaths|TestMapPathToFile' ./packages/cmd/` passes (`-vet=off` only to skip a pre-existing `go vet` finding in `run.go`, unrelated to this change)
- `infisical export --help` shows the new flag

The end-to-end fetch-and-write against a real project needs a live Infisical backend (the `test/` integration suite uses real credentials), so the resolution/mapping logic lives behind injectable helpers that are unit-tested directly.

One thing worth a maintainer's eye: I read "if the option is not provided, it will create a .env for all the paths" as a full recursive walk, and `apps/*` as a single level of children. Happy to adjust the glob/recursion semantics if you'd prefer different behavior (e.g. an explicit `--recursive`).

## Checklist

- [x] I have read the contributing guide and code of conduct

### DIFF
--- a/packages/cmd/export.go
+++ b/packages/cmd/export.go
@@ -92,6 +92,11 @@ var exportCmd = &cobra.Command{
 			util.HandleError(err, "Unable to parse flag")
 		}
 
+		toFiles, err := cmd.Flags().GetBool("to-files")
+		if err != nil {
+			util.HandleError(err, "Unable to parse flag")
+		}
+
 		request := models.GetAllSecretsParameters{
 			Environment:            environmentName,
 			TagSlugs:               tagSlugs,
@@ -128,6 +133,16 @@ var exportCmd = &cobra.Command{
 				util.HandleError(err)
 			}
 			util.PrintStdout(processedTemplate.String())
+			return
+		}
+
+		if toFiles {
+			if outputFile != "" {
+				util.PrintErrorMessageAndExit("--to-files cannot be combined with --output-file")
+			}
+			if err := runExportToFiles(request, format, tagSlugs, secretOverriding); err != nil {
+				util.HandleError(err, "Failed to export secrets to files")
+			}
 			return
 		}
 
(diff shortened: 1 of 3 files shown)


lollinng:
Thanks for the thorough review. Addressed both P1s and the redundant-MkdirAll P2 in f41ab80:

- **Path traversal**: server-returned folder names are now validated with `isSafeFolderSegment` before being mirrored into the filesystem; names that are `.`, `..`, empty, or contain a path separator are rejected with an error, so a malicious/buggy API response can't write outside the target dir.
- **Cycle detection**: `walkFolderTree` now carries a visited-set, so a cyclic folder graph terminates instead of recursing infinitely (added `TestExpandToFilePathsTerminatesOnCycle`).
- **Redundant MkdirAll**: dropped, since `writeToFile` already creates parent dirs.

On the P2 about a mid-pattern `*` (e.g. `apps/*/config`) being treated as a concrete path: that's intentional for this PR; the issue scopes the glob to a trailing `/*` (immediate children) plus the omitted-path recursive case. Happy to extend to deeper glob semantics if you'd like.

PR #279, opened 2026-06-24
fix(pam): bind and advertise local session proxies on loopback

Vligai:
## Context

The local PAM proxies for database, redis, and kubernetes sessions create their TCP listener with an
empty host (`":0"` / `":<port>"`), which Go resolves to all interfaces (`0.0.0.0` / `[::]`). These
proxies are consumed by the local client on the same machine, so they should only accept loopback
connections. This binds them to `127.0.0.1` and advertises `127.0.0.1` in the printed connection
string, so the client connects to exactly the address the proxy listens on. The rdp proxy already
binds and advertises `127.0.0.1`; this brings the other three in line.

Behavior: the proxy now accepts connections only from the local host. A client running on the same
machine connects via the printed `127.0.0.1` URL, as it already does for rdp sessions. Reaching the
proxy from another host, or from outside a container via a published port, no longer works (run the
client on the same machine). The printed connection string changes from `localhost:<port>` to
`127.0.0.1:<port>`.

- Before: listener on `:0` / `:<port>` (all interfaces); connection string used `localhost`.
- After: listener on `127.0.0.1:0` / `127.0.0.1:<port>` (loopback only); connection string uses
  `127.0.0.1`.

Changed files:

- `packages/pam/local/database-proxy.go`
- `packages/pam/local/redis-proxy.go`
- `packages/pam/local/kubernetes-proxy.go`
- `packages/pam/local/proxy_loopback_test.go` (test)

## Steps to verify the change

- Unit test (included in this PR): `go test ./packages/pam/local/ -run TestLocalProxiesBindLoopback`.
  It drives each proxy's `Start()` and asserts the listener binds a loopback address.
- Manual: start a database / redis / kubernetes PAM session and confirm the proxy port listens on
  `127.0.0.1` only (for example `lsof -iTCP -sTCP:LISTEN` or `netstat`), and that the local client
  still connects via the printed `127.0.0.1` URL.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Title follows the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) format: `type(scope): short description` (scope is optional, e.g., `fix: prevent crash on sync` or `fix(api): handle null response`).
- [x] Tested locally
- [ ] Updated docs (if needed)
- [ ] Updated CLAUDE.md files (if needed)
- [ ] Read the [contributing guide](https://infisical.com/docs/contributing/getting-started/overview)

### DIFF
--- a/packages/pam/local/database-proxy.go
+++ b/packages/pam/local/database-proxy.go
@@ -119,19 +119,19 @@ func StartDatabaseLocalProxy(accessToken string, accessParams PAMAccessParams, p
 
 	switch pamResponse.ResourceType {
 	case session.ResourceTypePostgres:
-		util.PrintfStderr("postgres://%s@localhost:%d/%s", username, proxy.port, database)
+		util.PrintfStderr("postgres://%s@127.0.0.1:%d/%s", username, proxy.port, database)
 	case session.ResourceTypeMysql:
-		util.PrintfStderr("mysql://%s@localhost:%d/%s", username, proxy.port, database)
+		util.PrintfStderr("mysql://%s@127.0.0.1:%d/%s", username, proxy.port, database)
 	case session.ResourceTypeMssql:
-		util.PrintfStderr("sqlserver://%s@localhost:%d?database=%s&encrypt=false&trustServerCertificate=true", username, proxy.port, database)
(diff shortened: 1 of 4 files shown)


Vligai:
merged into revamp branch instead.

PR #283, opened 2026-06-26
feat: switched secrets v3 endpoints to v4

akhilmhdh:
This PR moves the v3 endpoints to v4 for secrets commands. This also removed the personal override we are doing in CLI as v4 does it in server side. This allows imports to have overrides as well with the new behavior as fallback mechanism.


varonix0 (approved):

hotspoons:
FYI this PR broke older infisical clients and should have bumped the minor version - the patch version bump gave no indication of a breaking change and we needed to pin to the previous version

qbunt:
We also just had this break. Can I ask why breaking changes were made in a patch release? I also assumed that the patch version wouldn't have a major API version upgrade @varonix0

PR #307, opened 2026-07-09
feat(pam): azure cli access

x032205:
# Description 📣

Azure CLI access

Most of the logic is copied from the GCP account type (certain parts are mirrored 1:1)

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/packages/pam/handlers/azure/proxy.go
+++ b/packages/pam/handlers/azure/proxy.go
@@ -453,6 +453,7 @@ func isLoginHost(host string) bool {
 	return strings.EqualFold(host, "login.microsoftonline.com")
 }
 
+// key vault operations redact request bodies from logs since they often contain secrets
 func isVaultHost(host string) bool {
 	return strings.HasSuffix(strings.ToLower(host), ".vault.azure.net")
 }
--- a/packages/pam/handlers/azure/proxy.go
+++ b/packages/pam/handlers/azure/proxy.go
@@ -13,6 +13,7 @@ import (
 	"crypto/x509/pkix"
 	"encoding/base64"
 	"encoding/binary"
+	"encoding/json"
 	"encoding/pem"
 	"errors"
 	"fmt"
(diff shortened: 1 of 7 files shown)


packages/pam/handlers/azure/proxy.go
```diff
@@ -0,0 +1,506 @@
...
+	hostname := strings.Split(targetHost, ":")[0]
+	if !isAzureHost(hostname) {
+		writeErrorResponseTo(clientConn, fmt.Sprintf("host %q is not an allowed Azure endpoint", hostname))
+		return fmt.Errorf("rejected CONNECT to non-Azure host: %s", hostname)
+	}
+
+	_, err := clientConn.Write([]byte("HTTP/1.1 200 Connection Established\r\n\r\n"))
+	if err != nil {
+		return fmt.Errorf("failed to send CONNECT response: %w", err)
+	}
```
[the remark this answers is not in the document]
x032205: Azure endpoints will likely not ever redirect to a non-azure host unless the gateway machine is compromised

packages/pam/handlers/azure/proxy.go
```diff
@@ -0,0 +1,514 @@
...
+	sig := enc([]byte("infisical-pam-proxy-managed"))
+	return header + "." + payload + "." + sig
+}
+
+func isLoginHost(host string) bool {
+	return strings.EqualFold(host, "login.microsoftonline.com")
+}
+
+func isVaultHost(host string) bool {
+	return strings.HasSuffix(strings.ToLower(host), ".vault.azure.net")
```
bernie-g, line 458: This seems pretty random - why did we decide to do this only for this endpoint? I feel like this should be part of an actual log masking system that we implement rather than an ad hoc thing unless there's a good reason

x032205: It's because vault operations often return secret values so we outright disabled logging for them. I'll add that as a comment so future readers know

packages/pam/handlers/azure/proxy.go
```diff
@@ -0,0 +1,514 @@
...
+	if err != nil {
+		l.Error().Err(err).Str("url", targetURL.String()).Msg("Failed to forward request to Azure")
+		writeErrorResponseTo(writer, "failed to forward request to Azure")
+		return nil
+	}
+
+	resp.Header.Del("Connection")
+
+	var bodyCopy bytes.Buffer
+	if logRequest && !redactBody {
```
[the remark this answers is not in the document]
x032205: We'll leave this to session log masking

packages/pam/local/azure-proxy.go
```diff
@@ -0,0 +1,477 @@
...
+	fmt.Printf("\n")
+	if interactive {
+		fmt.Printf("  You are now in a shell wired to this session. Run 'az' as usual:\n")
+		util.PrintfStderr("    $ az account show\n")
+		util.PrintfStderr("    $ az group list\n")
+		fmt.Printf("\n")
+		fmt.Printf("  Type 'exit' to end the session.\n")
+	} else {
+		fmt.Printf("  Point the Azure CLI at this session, then run az commands:\n")
+		util.PrintfStderr("    $ export AZURE_CONFIG_DIR=\"%s\"\n", configDir)
```
bernie-g: I think we should print this both in interactive and non interactive sessions, otherwise people will be confused how to use this outside of the subshell

bernie-g (approved):

PR #309, opened 2026-07-11
feat: add WINRM logic

carlosmonastyrski:
# Description 📣

This PR adds support for delivering certificates to Windows machines over WinRM (Windows Remote Management). Until now, the Infisical gateway could reach Windows-based certificate authorities but couldn't push certificate files directly onto Windows hosts.

With this change, the gateway can securely connect to a Windows machine, verify it's reachable, write certificate files to it, and remove them when needed, all on behalf of Infisical's platform, which sits outside the customer's network and can't talk to those machines directly.

Connections are encrypted end to end (either through Windows' built-in message encryption over standard connections or through TLS for secured ones), and the target machine is always determined by Infisical's signed configuration rather than the request itself, so it can't be pointed at an unintended host.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


packages/gateway-v2/winrm_handler.go
```diff
@@ -0,0 +1,264 @@
...
+		}
+		files = append(files, winrm.FileDelivery{Path: f.Path, Content: content})
+	}
+	if err := winrm.DeliverFiles(ctx, credsFromEnv(ctx, env, p.winrmTransportParams), files); err != nil {
+		return nil, err
+	}
+	return map[string]any{"delivered": len(files)}, nil
+}
+
+func handleWinrmRemove(ctx context.Context, env *winrmRequestEnvelope) (any, error) {
```
sheensantoscapadngan: let's be very descriptive with the naming here please since `winrm` can be used for a lot of use-cases. from the current name, it's not obvious what's being removed. Same with the rest of the methods here

packages/gateway-v2/winrm_handler.go
```diff
@@ -0,0 +1,264 @@
...
+	log.Debug().Int("status", rw.status).Msg("winrm: response written")
+	return nil
+}
+
+var serveWinrmMux = sync.OnceValue(func() *http.ServeMux {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/v1/test", wrapWinrm(handleWinrmTest))
+	mux.HandleFunc("/v1/deliver", wrapWinrm(handleWinrmDeliver))
+	mux.HandleFunc("/v1/remove", wrapWinrm(handleWinrmRemove))
+	return mux
```
sheensantoscapadngan, line 135: same remark here regarding the func endpoints

packages/gateway-v2/winrm/winrm.go
```diff
@@ -0,0 +1,336 @@
...
+// newClient builds a WinRM client. HTTP (the default) uses NTLM message encryption to keep the SOAP
+// body confidential without a server certificate; HTTPS uses TLS, verifying the listener against the
+// system trust store, an optional pinned CA (for a self-signed listener), or skipping verification
+// when Insecure is set.
+func newClient(ctx context.Context, creds Credentials) (*winrm.Client, error) {
+	params := *winrm.DefaultParameters
+	if creds.UseHTTPS {
+		// The bounded dial caps the response read; the library otherwise reads the body unbounded.
+		params.TransportDecorator = func() winrm.Transporter { return winrm.NewClientNTLMWithDial(boundedDial(ctx)) }
+	} else {
```
sheensantoscapadngan, line 221: wait is this intentional? when HTTPS is to be used, we use NTLM?

packages/gateway-v2/winrm/winrm.go
```diff
@@ -0,0 +1,336 @@
...
+		enc, err := winrm.NewEncryption("ntlm")
+		if err != nil {
+			return nil, fmt.Errorf("%w: failed to initialize NTLM message encryption: %v", ErrConnect, err)
+		}
+		params.TransportDecorator = func() winrm.Transporter { return enc }
+	}
+
```
sheensantoscapadngan, line 234: IGNORE PLACEMENT

also seems like we're missing response capping for http path?

packages/gateway-v2/winrm/winrm.go
carlosmonastyrski: Intentional, NTLM here is the authentication, not the encryption. Over HTTPS, TLS encrypts the channel and NTLM only authenticates the session (NewClientNTLMWithDial). Over HTTP there’s no TLS, so we use NewEncryption("ntlm"), which additionally NTLM-seals the SOAP body so the certificate and key stay confidential (default WinRM listeners require this on HTTP). Sealing over HTTPS would be redundant double-encryption and isn’t what HTTPS listeners expect. I added a comment making the auth-vs-encryption distinction clear.

packages/gateway-v2/winrm/winrm.go
carlosmonastyrski: You’re right, only HTTPS was capped. The HTTP transport (masterzen’s NTLM-sealing Encryption) uses a bare &http.Client{} that falls back to Go’s global http.DefaultTransport and gives us no hook to replace it (the HTTPS path does, which is why that one was already capped). So at gateway startup we wrap http.DefaultTransport with a RoundTripper that caps each response body at 32 MB (same as HTTPS) and delegates the rest. It’s scoped to just this path, the gateway’s other clients, all use their own transports, and only trips on an abusive endless body, never on real (tiny) command responses.

packages/gateway-v2/winrm_handler.go
sheensantoscapadngan: doesn't seem like you updated `/v1/test`, `/v1/deliver`, etc. 
image

packages/gateway-v2/gateway.go
```diff
@@ -443,6 +444,9 @@ func (g *Gateway) registerHeartBeat(ctx context.Context, errCh chan error) {
 func (g *Gateway) Start(ctx context.Context) error {
 	log.Info().Msgf("Starting gateway")
 
+	// Cap WinRM HTTP-transport response bodies.
+	winrm.InstallHTTPResponseCap()
+
```
sheensantoscapadngan, line 449: this modifies it for the entire gateway though right? is there no way for us to scope it to just the ones in the winRM path? seems off to apply this generally since we might have valid use-cases for large responses

packages/gateway-v2/winrm/winrm.go
```diff
@@ -0,0 +1,238 @@
...
+		creds.UseHTTPS,
+		creds.Insecure,
+		nil, nil, nil,
+		commandDeadline,
+	)
+	client, err := winrm.NewClientWithParameters(endpoint, creds.Username, creds.Password, &params)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %v", ErrConnect, err)
+	}
+	return client, nil
```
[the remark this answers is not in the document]
sheensantoscapadngan: how did we address this?

packages/gateway-v2/gateway.go
carlosmonastyrski: It does replace the process `http.DefaultTransport`, but the wrapper only caps responses whose path is `/wsman` (WinRM's WS-Management endpoint) and passes everything else through untouched, so large responses elsewhere aren't affected. The reason it can't be scoped at the client level is that masterzen's NTLM message-encryption transport uses a bare `*http.Client` with no hook to inject a custom transport, so `DefaultTransport` is the only seam without forking the library. The HTTPS path is capped separately via `boundedDial`.

sheensantoscapadngan (approved):

PR #323, opened 2026-07-21
feat: add relay systemd install with enroll method

sheensantoscapadngan:
# Description 📣

This PR adds enrollment-based systemd installation for relays. The main changes are:
- Token enrollment before installing an org relay service.
- AWS authentication configuration for service startup.
- Positional relay names with legacy install compatibility.
- New relay environment keys and systemd configuration writers.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


packages/relay/systemd.go
```diff
@@ -81,6 +81,127 @@ func InstallRelaySystemdService(token string, domain string, name string, host s
...
+	if domain != "" {
+		configContent += fmt.Sprintf("INFISICAL_API_URL=%s\n", domain)
+	}
+
+	environmentFilePath := filepath.Join(configDir, "relay.conf")
+	if err := os.WriteFile(environmentFilePath, []byte(configContent), 0600); err != nil {
+		return fmt.Errorf("failed to write environment file: %v", err)
+	}
+
+	serviceName := "infisical-relay"
```
bernie-g: We're hardcoding the relay to be the same name whereas on the gateway we set the service name to be whatever the user entered. This also makes it so you can only have one systemd relay service running on the machine

packages/cmd/relay.go
```diff
@@ -111,44 +111,53 @@ var relayStartCmd = &cobra.Command{
...
-			storedEnrollToken, _ := relay.LoadStoredEnrollmentToken(relayName)
-			alreadyEnrolled := storedEnrollToken != "" && storedEnrollToken == enrollToken
-
-			if alreadyEnrolled {
-				log.Info().Msg("Enrollment token matches stored token. Skipping enrollment.")
+			// Check if we already have a stored access token (e.g., from systemd install)
+			storedAccessToken, _ := relay.LoadStoredAccessToken(relayName)
+			if storedAccessToken != "" {
+				// Already have a valid access token, use it directly
+				enrolledAccessToken = storedAccessToken
```
bernie-g: Wouldn't this make it so --token is useless if you have a stored token already? Pretty sure that in the gateway version of this if you pass in --token it overrides the stored token

bernie-g (changes requested):

packages/cmd/relay.go
```diff
@@ -111,24 +111,15 @@ var relayStartCmd = &cobra.Command{
...
-				log.Info().Msg("Using stored access token...")
-			} else {
-				// No stored token, require enrollment token to enroll
-				enrollToken, _ := cmd.Flags().GetString("token")
-				if enrollToken == "" {
-					util.HandleError(errors.New("--token is required when --enroll-method=token"))
-				}
+			enrollToken, _ := cmd.Flags().GetString("token")
 
+			if enrollToken != "" {
```
bernie-g, line 116: ignore comment placement

We're missing a couple of edge cases for installing a named relay service when an existing legacy service already exists. Look at `detectLegacyService` and `resolveInstallPaths` in `/gateway-v2/systemd.go`. That'll handle a few possible collisions with legacy relays

bernie-g (changes requested):

bernie-g (approved):

PR #328, opened 2026-07-22
feat(pam): conn test for other accs

x032205:
# Description 📣

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


packages/gateway-v2/test_connection_handler.go
```diff
@@ -121,6 +301,18 @@ func doTCPReachabilityTest(ctx context.Context, host string, port int) error {
...
+// runWithContext bounds op by ctx even when the underlying driver ignores context cancellation
+func runWithContext(ctx context.Context, op func() error) error {
+	done := make(chan error, 1)
+	go func() { done <- op() }()
+	select {
+	case err := <-done:
+		return err
+	case <-ctx.Done():
+		return fmt.Errorf("connection test timed out")
+	}
```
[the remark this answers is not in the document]
x032205: False alarm

packages/gateway-v2/test_connection_handler.go
```diff
@@ -148,38 +340,65 @@ func handleTestConnection(w http.ResponseWriter, r *http.Request) {
...
+				Certificate: params.Certificate,
+				TimeoutMs:   env.TimeoutMs,
+			})
+			return err
+		}
 	case testConnModeTCP:
-		testErr = doTCPReachabilityTest(ctx, target.host, target.port)
+		op = func() error { return doTCPReachabilityTest(ctx, target.host, target.port) }
 	default:
 		writeRPCError(w, http.StatusBadRequest, fmt.Sprintf("unsupported test-connection mode: %q", env.Mode))
```
bernie-g, line 486: Wouldn't this brick all old gateways that still use postgres/mysql/etc? Since we're switching to just "sql"

x032205: Yes, there's I think ~3 minor releases that expect postgres. Mysql and the others should not be a concern

packages/gateway-v2/test_connection_handler.go
```diff
@@ -72,42 +109,185 @@ type testConnectionResponse struct {
...
+
+// doKubernetesConnectionTest confirms the API server is reachable and accepts the token (401 = bad credentials)
+func doKubernetesConnectionTest(ctx context.Context, host string, port int, params kubernetesTestParams) error {
+	tlsConfig, err := buildTestTLSConfig(host, params.SslCertificate, params.SslRejectUnauthorized)
+	if err != nil {
+		return err
+	}
+
+	client := &http.Client{Transport: &http.Transport{TLSClientConfig: tlsConfig}}
+	url := fmt.Sprintf("https://%s/api", net.JoinHostPort(host, strconv.Itoa(port)))
```
saifsmailbox98, line 279: what about `http` ones?

packages/gateway-v2/test_connection_handler.go
```diff
@@ -72,42 +109,185 @@ type testConnectionResponse struct {
...
+	defer db.Close()
 
 	var result int
-	return conn.QueryRow(ctx, "SELECT 1").Scan(&result)
+	return db.QueryRowContext(ctx, "SELECT 1").Scan(&result)
+}
+
+// doMongoConnectionTest authenticates against the target MongoDB and pings it
+func doMongoConnectionTest(ctx context.Context, host string, port int, params mongoTestParams) error {
+	opts := options.Client().SetHosts([]string{net.JoinHostPort(host, strconv.Itoa(port))}).SetDirect(true)
```
saifsmailbox98, line 210: Can we pass the actual connection string to the gateway so that multihost, etc. are validated correctly? We also don't need to do SRV resolution on the backend.

saifsmailbox98 (changes requested):

packages/gateway-v2/test_connection_handler.go
x032205: I think we can just force https like this because 99% of cases http would be a misconfiguration. if a customer complains then we can later accommodate

packages/gateway-v2/test_connection_handler.go
x032205: I think for now this is fine. The idea here is to be as secure as possible and prevent SSRF.

packages/gateway-v2/test_connection_handler.go
```diff
@@ -72,42 +115,268 @@ type testConnectionResponse struct {
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusUnauthorized {
+		return fmt.Errorf("kubernetes API rejected the gateway's pod token (HTTP %d)", resp.StatusCode)
+	}
+	if resp.StatusCode == http.StatusForbidden {
```
saifsmailbox98, line 373: this passes even if the service account doesn't exist

better if we get them using the following:
```
GET /api/v1/namespaces/{namespace}
GET /api/v1/namespaces/{namespace}/serviceaccounts/{serviceAccountName}
```

feel free to ignore it as well, can be done later

saifsmailbox98 (approved):

PR #338, opened 2026-07-28
feat(pam): redis account type

x032205:
# Description 📣

Redis account type

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation


packages/pam/local/access.go
```diff
@@ -381,6 +381,84 @@ func startDatabaseProxy(httpClient *resty.Client, response *api.PAMAccessRespons
...
+	}
+	fmt.Printf("  Account:   %s\n", account)
+	fmt.Printf("  Duration:  %s\n", duration.String())
+	fmt.Printf("\n")
+	fmt.Printf("  Host:      127.0.0.1\n")
+	fmt.Printf("  Port:      %d\n", port)
+	if username != "" {
+		fmt.Printf("  Username:  %s\n", username)
+	}
+	fmt.Printf("  Password:  (not required)\n")
```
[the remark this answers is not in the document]
x032205: not an issue specific to redis. all types share this logic. we should fix this in a different PR

packages/pam/handlers/redis/proxy.go
```diff
@@ -69,9 +69,15 @@ func (p *RedisProxy) HandleConnection(ctx context.Context, clientConn net.Conn)
...
 	// Only authenticate if credentials are provided
-	if p.config.InjectUsername != "" && p.config.InjectPassword != "" {
-		if err := selfToClientRedisConn.Writer().WriteCommand("AUTH", p.config.InjectUsername, p.config.InjectPassword); err != nil {
-			return err
+	if p.config.InjectPassword != "" {
+		var writeErr error
+		if p.config.InjectUsername != "" {
+			writeErr = selfToClientRedisConn.Writer().WriteCommand("AUTH", p.config.InjectUsername, p.config.InjectPassword)
+		} else {
+			writeErr = selfToClientRedisConn.Writer().WriteCommand("AUTH", p.config.InjectPassword)
```
saifsmailbox98, line 77: This can never run. Username is always set so we always take the 2-arg branch above and not this.


This may also break redis 5 since it only takes one arg. should we check for `!= "default"` instead? (Double check the compatibility with the newer versions once if one-arg is still supported by them for default)

packages/gateway-v2/test_connection_handler.go
```diff
@@ -232,6 +242,76 @@ func doMongoConnectionTest(ctx context.Context, host string, port int, params mo
...
+	reader := resp3.NewReader(conn)
+	writer := resp3.NewWriter(conn)
+
+	// only authenticate when password is supplied
+	if params.Password != "" {
+		var err error
+		if params.Username != "" {
+			err = writer.WriteCommand("AUTH", params.Username, params.Password)
+		} else {
+			err = writer.WriteCommand("AUTH", params.Password)
```
saifsmailbox98, line 285: same comment as the other one

packages/pam/local/access.go
```diff
@@ -107,7 +107,7 @@ func StartPAMAccess(accessToken, path, reason, durationStr, targetHost string, p
 	case AccountTypeSSH:
 		startSSHAccess(httpClient, &pamResponse, displayPath, durationStr, port)
 	case AccountTypeRedis:
-		util.PrintErrorMessageAndExit("Redis access not yet supported in the new PAM model")
+		startRedisProxy(httpClient, &pamResponse, displayPath, durationStr, port)
```
saifsmailbox98, line 110: it has no usages, can we delete it? and also HandleApprovalWorkflow and PAMAccessParams

packages/pam/local/access.go
```diff
@@ -381,6 +381,84 @@ func startDatabaseProxy(httpClient *resty.Client, response *api.PAMAccessRespons
...
+	go func() {
+		sig := <-sigChan
+		log.Info().Msgf("Received signal %v, initiating graceful shutdown...", sig)
+		proxy.gracefulShutdown()
+	}()
+
+	proxy.Run()
+}
+
+func printRedisSessionInfo(folder, account string, duration time.Duration, username string, port int) {
```
saifsmailbox98, line 437: let's make it consistent with others. it is missing the Connection Details/How to Connect dividers


Does redis support connection string as documented here? `redis://127.0.0.1:<port>`

https://infisical-pam-358.mintlify.site/cli/commands/pam#databases-postgresql-mysql-sql-server-mongodb-redis

If so, let's print that as well

saifsmailbox98 (changes requested):

packages/gateway-v2/test_connection_handler.go
```diff
@@ -25,6 +25,7 @@ import (
 	"github.com/jackc/pgx/v5/stdlib"
```
saifsmailbox98, line 25: IGNORE COMMENT PLACEMENT

Merge main to see if the go vuln scan check passes

saifsmailbox98 (changes requested):

saifsmailbox98 (approved):

PR #343, opened 2026-07-31
feat: add Gateway support for windows commands

carlosmonastyrski:
# Description 📣

Adds a `POST /v1/run-command` WinRM endpoint so the control plane can run one operator-supplied command on a Windows host after a certificate sync delivers its files.

The command is wrapped in a PowerShell script that reports its own outcome in a trailer tagged with a per-run nonce, rather than relying on the process exit code. Measured on a real host, an `exit` inside the wrapper is swallowed and arrives as 0, so a failed command would otherwise be reported as a successful sync. The script uses `$?` rather than `$LASTEXITCODE` so the outcome is the last statement's, matching how `sh -c` decides it on Linux; reading `$LASTEXITCODE` alone fails a run where `findstr` or `robocopy` exited non-zero routinely.

Output is capped per stream, with the last bytes retained so a command that floods stdout and then fails is still reported as failed. PowerShell's CLIXML stderr is decoded to plain text. A requested timeout above the ceiling is clamped rather than reset to the default.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


packages/gateway-v2/winrm_handler.go
```diff
@@ -292,6 +311,43 @@ func handleWinrmRemoveFiles(ctx context.Context, env *winrmRequestEnvelope) (any
...
 }
 
+// handleWinrmRunCommand runs a single command on the host. A non-zero exit code is not an RPC error
+func handleWinrmRunCommand(ctx context.Context, env *winrmRequestEnvelope) (any, error) {
+	var p winrmRunCommandParams
+	if err := json.Unmarshal(env.Params, &p); err != nil {
+		return nil, fmt.Errorf("malformed run-command params")
+	}
+	if strings.TrimSpace(p.Command) == "" {
+		return nil, fmt.Errorf("command is required")
```
bernie-g, line 321: Not an issue with this PR specifically but I noticed we're missing a `forwardConfig.ActorType != ActorTypePlatform` check in this if branch in gateway.go. Can we add that here?

Image

bernie-g (approved):

PR #345, opened 2026-07-31
fix(auth): preserve the session domain during automatic re-login

### LINKED ISSUE #312: Auto re-login drops --domain and ignores INFISICAL_API_URL for EU Cloud — expired EU sessions open the US login page and silently flip LoggedInUserDomain to US
### Summary

When a logged-in user's session JWT expires, the CLI's automatic re-login flow cannot be pinned to EU Cloud: it drops the parent command's `--domain` flag, deliberately ignores `INFISICAL_API_URL` when it points at a default cloud URL, defaults its region prompt to the US, and opens a browser at `https://app.infisical.com/login`. If that US login is completed, the config file's `LoggedInUserDomain` is silently rewritten to US, breaking every subsequent command for an EU-hosted org.

For EU Cloud customers running unattended scripts (hooks, cron/launchd jobs), this manifests as recurring, apparently random US-login browser pop-ups that no amount of domain configuration prevents.

### Version / platform

- CLI 0.43.89 (Homebrew, macOS). Code paths verified identical in the `v0.43.105` tag.
- Account/org on Infisical Cloud EU (`https://eu.infisical.com`).

### Steps to reproduce

1. Log in to EU Cloud: `infisical login --domain https://eu.infisical.com` (browser flow). `~/.infisical/infisical-config.json` correctly records `LoggedInUserDomain: https://eu.infisical.com/api`.
2. Let the session JWT expire.
3. Run any user-auth command with the domain explicitly pinned, e.g.
   `infisical secrets get FOO --projectId <id> --env=dev --plain --silent --domain https://eu.infisical.com`
   (or with `INFISICAL_API_URL=https://eu.infisical.com` exported).

### Actual behaviour

- `util.GetAllEnvironmentVariables` detects the expired token and calls `EstablishUserLoginSession()` (`packages/util/secrets.go:307-309`).
- `EstablishUserLoginSession()` re-execs the CLI as `exec.Command(exePath, "login", "--silent")` (`packages/util/auth.go:74`) — **only `--silent` is forwarded; the parent's `--domain` is dropped**.
- In the child, `usePresetDomain()` returns false even though `INFISICAL_API_URL` is inherited, because default cloud URLs are explicitly excluded unless `--domain` was set on that invocation (`packages/cmd/login.go:465`):
  ```go
  shouldUsePresetDomain := preconfiguredUrl != "" && (domainFlagExplicitlySet ||
      (preconfiguredUrl != util.INFISICAL_DEFAULT_US_URL && preconfiguredUrl != util.INFISICAL_DEFAULT_EU_URL))
  ```
- Control falls to `askForDomain()` (`packages/cmd/login.go:503`), whose default-highlighted option is **"Infisical Cloud (US Region)"**.
- `browserCliLogin()` opens the browser at `config.INFISICAL_LOGIN_URL` → `https://app.infisical.com/login?...` (`packages/cmd/login.go:873-896`). Note `browser.OpenURL` runs before the TTY check, so semi-interactive contexts (editor-spawned hooks etc.) still get the browser window.
- If the user completes the US login, `WriteInitalConfig` persists `LoggedInUserDomain: config.INFISICAL_URL` (`packages/util/config.go:55`) — the stored EU domain is silently replaced with US.

### Expected behaviour

- The automatic re-login should reuse the domain of the session that expired (`configFile.LoggedInUserDomain`), or at minimum honor the parent command's `--domain` / `INFISICAL_DOMAIN` / `INFISICAL_API_URL`.
- An EU Cloud value in `INFISICAL_API_URL` should not be discarded by `usePresetDomain()`.
- A completed re-login should not silently change the stored account domain.

### Suggested fixes

1. In `EstablishUserLoginSession()`, forward the domain: `exec.Command(exePath, "login", "--silent", "--domain", <current config.INFISICAL_URL or LoggedInUserDomain>)`.
2. In `usePresetDomain()`, accept a preconfigured EU/US cloud URL when it matches the stored `LoggedInUserDomain`.
3. Guard `askForDomain()`/`browserCliLogin()` behind a TTY check so unattended runs fail loudly instead of opening a browser.

### Related

- Infisical/infisical#5051 fixed `--domain`/`INFISICAL_API_URL` for *explicit* `infisical login` invocations, but not this auto-re-login child process, and the EU-cloud exclusion in `usePresetDomain` remains.
- #94 (no token auto-refresh) makes this path frequent for user-auth scripts.

### Workarounds found

- Re-login manually with the explicit flag: `infisical login --domain https://eu.infisical.com`.
- For unattended scripts: use machine identity / `--token` (universal auth), which never reaches the interactive login path.

BenyD:
# Description 📣

`EstablishUserLoginSession()` re-execs the CLI as `login --silent`, forwarding only `--silent`. The child then resolves its domain from scratch, and because `usePresetDomain()` deliberately ignores default cloud URLs unless `--domain` was set on that specific invocation, an expired EU Cloud session falls through to the region prompt, opens `https://app.infisical.com/login`, and on completion rewrites `LoggedInUserDomain` to US. Every subsequent command then targets the wrong region.

This forwards the domain the parent has already resolved.

**Why this is the only change needed.** Because the child now sees an explicit `--domain`, `cmd.Flags().Changed("domain")` is true, so `usePresetDomain()` accepts the value even for EU/US cloud URLs. The exclusion at `login.go:465` is never reached, which means no change to `usePresetDomain()` itself. That function is shared with the normal `infisical login` path, so leaving its semantics alone seemed preferable to widening them. This also fixes every caller of `EstablishUserLoginSession()` at once rather than per command.

**Precedence.** The stored `LoggedInUserDomain` wins, since it is the instance whose session just expired. Otherwise `config.INFISICAL_URL` is used, as already resolved by the root command from `--domain`, `INFISICAL_DOMAIN`, or `.infisical.json`.

**One deliberate exception:** the bare US default is not forwarded. `config.INFISICAL_URL` defaults to the US URL rather than empty, so forwarding it unconditionally would set an explicit `--domain` on every re-login and silently skip the region prompt for users who have never configured one. Returning `""` in that case preserves today's behaviour exactly.

**Scoped to one of the three fixes suggested in #312.** Suggestion 2 is subsumed by the above. Suggestion 3, guarding `askForDomain()`/`browserCliLogin()` behind a TTY check, is a UX change for semi-interactive contexts and felt like a maintainer's call rather than something to fold into a bug fix. Happy to add it or open it separately if you want it.

Fixes #312

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Split the decision into two pure functions, `resolveReLoginDomain` and `buildReLoginArgs`, so the logic is testable without spawning a process. This mirrors the existing `GetUpdateInstructions()` / `getUpdateInstructions(goos, execPath)` split in `check-for-update.go`.

New `packages/util/auth_test.go` covers stored EU domain, stored self-hosted domain, explicit non-default override with nothing stored, the bare US default, nothing known at all, and stored US Cloud.

```sh
go test ./packages/util/ -run 'TestResolveReLoginDomain|TestBuildReLoginArgs' -v
```

```
--- PASS: TestResolveReLoginDomain (0.00s)
    --- PASS: .../stored_EU_domain_wins_over_US_default
    --- PASS: .../stored_self-hosted_domain_is_forwarded
    --- PASS: .../explicit_non-default_resolved_URL_is_forwarded_when_nothing_is_stored
    --- PASS: .../bare_US_default_is_not_forwarded
    --- PASS: .../nothing_known_at_all
    --- PASS: .../stored_US_domain_is_still_forwarded
--- PASS: TestBuildReLoginArgs (0.00s)
```

Verified the tests are meaningful by forcing `resolveReLoginDomain` to always return `""`, reproducing the pre-fix behaviour: the EU, self-hosted, and explicit-override cases all fail. Full `packages/util` suite passes with no regressions, and `go build ./...` is clean.

**What I could not test:** I do not have an EU Cloud account, so this was diagnosed by reading the code paths rather than reproducing end to end. The unit tests cover the decision logic; a reviewer with EU access confirming the real flow would be worthwhile.

---

### DIFF
--- a/packages/util/auth.go
+++ b/packages/util/auth.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/exec"
 
+	"github.com/Infisical/infisical-merge/packages/config"
 	infisicalSdk "github.com/infisical/go-sdk"
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
@@ -60,6 +61,35 @@ func IsAuthMethodValid(authMethod string, allowUserAuth bool) (isValid bool, str
 	return false, ""
 }
 
+// resolveReLoginDomain decides which domain the automatic re-login child process
+// should be pinned to, or returns "" to leave the child's normal domain
+// resolution (and its interactive region prompt) untouched.
+//
+// storedDomain is LoggedInUserDomain from the config file, i.e. the instance
+// whose session just expired, so it wins. resolvedURL is config.INFISICAL_URL as
+// already resolved by the root command from --domain, INFISICAL_DOMAIN or
+// .infisical.json; it is only forwarded when it differs from the built-in US
+// default, since that default is what an unconfigured invocation looks like and
+// forwarding it would silently skip the region prompt for first-time logins.
+func resolveReLoginDomain(storedDomain string, resolvedURL string) string {
+	if storedDomain != "" {
+		return storedDomain
+	}
+	if resolvedURL != "" && resolvedURL != AppendAPIEndpoint(INFISICAL_DEFAULT_US_URL) {
+		return resolvedURL
+	}
+	return ""
+}
+
+// buildReLoginArgs returns the argv for the re-login child process.
+func buildReLoginArgs(domain string) []string {
+	args := []string{"login", "--silent"}
+	if domain != "" {
+		args = append(args, "--domain", domain)
+	}
+	return args
+}
+
 // EstablishUserLoginSession handles the login flow to either create a new session or restore an expired one.
 // It returns fresh user details if login is successful.
 func EstablishUserLoginSession() LoggedInUserDetails {
@@ -70,8 +100,18 @@ func EstablishUserLoginSession() LoggedInUserDetails {
 		PrintErrorMessageAndExit(fmt.Sprintf("Failed to determine executable path: %v", err))
 	}
 
+	// The child process resolves its own domain from scratch and would otherwise
+	// fall back to US Cloud, so forward the domain this session was using.
+	// Without this an expired EU Cloud session re-prompts for a region, opens the
+	// US login page, and on completion rewrites LoggedInUserDomain to US.
+	var storedDomain string
+	if configFile, configErr := GetConfigFile(); configErr == nil {
+		storedDomain = configFile.LoggedInUserDomain
+	}
+	domain := resolveReLoginDomain(storedDomain, config.INFISICAL_URL)
+
 	// Spawn infisical login command
-	loginCmd := exec.Command(exePath, "login", "--silent")
+	loginCmd := exec.Command(exePath, buildReLoginArgs(domain)...)
 	loginCmd.Stdin = os.Stdin
 	loginCmd.Stdout = os.Stdout
 	loginCmd.Stderr = os.Stderr
--- a/packages/util/auth_test.go
+++ b/packages/util/auth_test.go
@@ -1,0 +1,98 @@
+package util
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestResolveReLoginDomain(t *testing.T) {
+	usDefault := AppendAPIEndpoint(INFISICAL_DEFAULT_US_URL)
+	euAPI := AppendAPIEndpoint(INFISICAL_DEFAULT_EU_URL)
+
+	tests := []struct {
+		name         string
+		storedDomain string
+		resolvedURL  string
+		expected     string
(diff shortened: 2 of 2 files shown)


BenyD:
Gentle bump on this one. It is still mergeable with no conflicts.

If the scope feels large for a fix, I am glad to trim it down or add a test case for the re-login path. Happy to rebase onto current main as well if that helps.

PR #351, opened 2026-08-03
feat: add pam agent

sheensantoscapadngan:
# Description 📣
This PR adds support for PAM agent usage

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


packages/cmd/pam.go
```diff
@@ -68,12 +76,285 @@ The path format is: /folder/account-name (leading slash optional)`,
...
+
+Sessions are created lazily: binding the ports costs nothing, and an account the agent
+never connects to never opens a PAM session.
+
+Accounts behind an approval policy are included, and their approval requests are raised
+lazily too: the first time the agent touches such an account, a request is raised for it
+and that attempt fails. Once a reviewer approves, the next attempt works, with nothing to
+restart. An account the agent never touches never puts a request in front of anyone. Pass
+--no-approval-request to leave these accounts out entirely.
+
```
sheensantoscapadngan: too much technical details no?

packages/cmd/pam.go
```diff
@@ -68,12 +76,285 @@ The path format is: /folder/account-name (leading slash optional)`,
...
+Per-account guidance for the agent comes from the account's description in Infisical.
```
sheensantoscapadngan: not relevant anymore

packages/cmd/pam.go
```diff
@@ -68,12 +76,285 @@ The path format is: /folder/account-name (leading slash optional)`,
...
+
+Instructions are delivered in whatever format the agent understands, and their path is
+always exported as INFISICAL_PAM_CONTEXT_FILE so custom agents can pick them up too.
+
+Authenticates as the logged-in user by default, or as a machine identity when --token
+or INFISICAL_UNIVERSAL_AUTH_ACCESS_TOKEN is set.`,
```
sheensantoscapadngan: no need to mention this anymore

packages/cmd/pam.go
```diff
@@ -68,12 +76,285 @@ The path format is: /folder/account-name (leading slash optional)`,
...
+// agenticAuthMethods lists what --auth-method accepts, for the flag help and for the error when it
+// doesn't. Kept next to the strategy table below so the two cannot drift apart.
+const agenticAuthMethods = "universal-auth, kubernetes, azure, gcp-id-token, gcp-iam, aws-iam, oidc-auth, jwt-auth, ldap-auth"
+
+// authenticateMachineIdentity exchanges a machine identity's own credentials for an access token when
+// --auth-method (or INFISICAL_AUTH_METHOD) names one. It returns "" when no method was asked for, which
+// leaves the caller on its stored-login path.
+//
+// This keeps its own strategy table rather than sharing one with login, gateway and kmip, which each
+// keep theirs. Consolidating the four is worth doing, but not from here.
```
sheensantoscapadngan: unnecessary detail

packages/cmd/pam.go
```diff
@@ -68,12 +76,285 @@ The path format is: /folder/account-name (leading slash optional)`,
...
+	if err != nil {
+		util.HandleError(err, "Unable to get custom headers")
+	}
+
+	client := infisicalSdk.NewInfisicalClient(context.Background(), infisicalSdk.Config{
+		SiteUrl:          config.INFISICAL_URL,
+		UserAgent:        api.USER_AGENT,
+		AutoTokenRefresh: false,
+		CustomHeaders:    customHeaders,
+	})
```
sheensantoscapadngan: why is AutoTokenRefresh false? expectation here is for this to be a long-running process in the remote agent use-case...

packages/pam/agent/adapters.go
```diff
@@ -0,0 +1,253 @@
...
+	// useSystemPromptFlag appends the document as a CLI flag instead of writing a file.
+	useSystemPromptFlag bool
+}
+
+var adapters = []Adapter{
+	// Claude Code takes instructions on the command line, so nothing needs to touch the repo.
+	{Name: "claude", useSystemPromptFlag: true},
+	{Name: "codex", contextFile: "AGENTS.md"},
+	{Name: "gemini", contextFile: "GEMINI.md"},
+}
```
sheensantoscapadngan, line 40: does this overwrite existing files? if so, is that the most ideal approach?

packages/pam/agent/run.go
```diff
@@ -0,0 +1,560 @@
...
+	}
+	return path, nil
+}
+
+// kubeContextName turns an account path into a context name.
+//
+// The name reaches kubectl and the instruction document, so it is reduced to letters, digits, dots,
+// dashes and underscores. Names also have to stay distinct: contexts live in a map, so a repeat would
+// overwrite an entry and leave one cluster answering on another one's proxy.
+func kubeContextName(path string, taken map[string]bool) string {
```
sheensantoscapadngan: it might be cleaner to have a separate file altogether for the kubernetes-related methods

packages/pam/agent/instructions.go
```diff
@@ -0,0 +1,199 @@
...
+	case "postgres":
+		return fmt.Sprintf("postgres://%s@127.0.0.1:%d/%s", placeholderUser, port, placeholderDatabase),
+			fmt.Sprintf("psql -h 127.0.0.1 -p %d -U %s -d %s", port, placeholderUser, placeholderDatabase)
+	case "mysql":
+		return fmt.Sprintf("mysql://%s@127.0.0.1:%d", placeholderUser, port),
+			fmt.Sprintf("mysql -h 127.0.0.1 -P %d -u %s", port, placeholderUser)
+	case "mssql":
+		return fmt.Sprintf("sqlserver://%s@127.0.0.1:%d", placeholderUser, port),
+			fmt.Sprintf("sqlcmd -S 127.0.0.1,%d -U %s", port, placeholderUser)
+	case "mongodb":
```
saifsmailbox98: the database is not mentioned, so mongosh connects to the "test" db, and base command times out, need to add "directConnection=true"


can we use what we get back from session start command for each?

packages/pam/agent/run.go
```diff
@@ -0,0 +1,528 @@
...
+}
+
+// infisicalAuthEnvKeys are the variables the CLI's own auth resolution reads. The agent reaches its
+// accounts over loopback and never needs an Infisical credential, so none of these are handed down:
+// one that was would let it call the API directly and open sessions outside the accounts, duration and
+// approval gates this run was launched with.
+//
+// This is hygiene, not a boundary. A child running as the same user can still read the parent's argv,
+// so it is the sandbox below, not this list, that contains a deliberately hostile agent.
+var infisicalAuthEnvKeys = []string{
```
saifsmailbox98, line 182: DOCKER_HOST/CONTAINER_HOST/DOCKER_CONTEXT are not stripped

packages/pam/agent/run.go
```diff
@@ -0,0 +1,528 @@
...
+// every mask in pamDenyPaths, and then authenticate as the caller to open PAM sessions outside this
+// run's accounts, duration and approval gates.
+//
+// These are kept apart from pamDenyPaths because a path mask does not cover a socket. Connecting to one
+// is a network operation, so it survives a file deny and needs the socket-specific control instead.
+//
+// Symlinked sockets are listed by their target as well, since seatbelt matches the resolved path and
+// Docker Desktop points /var/run/docker.sock at the one under ~/.docker.
+func pamDenySockets(home string) []string {
+	candidates := []string{
```
saifsmailbox98, line 337: doesn't work on linux currently

packages/cmd/pam.go
```diff
@@ -68,12 +76,287 @@ The path format is: /folder/account-name (leading slash optional)`,
...
+		}
+		duration, err := time.ParseDuration(durationStr)
+		if err != nil {
+			util.PrintErrorMessageAndExit(fmt.Sprintf("Invalid duration %q. Use formats like '1h', '30m', '2h30m'.", durationStr))
+		}
+		if duration <= 0 {
+			util.PrintErrorMessageAndExit(fmt.Sprintf("Duration %q must be positive.", durationStr))
+		}
+
+		accounts = splitAccountFlags(accounts)
```
saifsmailbox98: `--account ","`, `--account ,`, `--account " "` proxies the entire org

also not mentioning `--account` proxies all accounts, right? do you think it's the correct default?

can we by default not proxy for any and ask for atleast one `--account`, and only proxy all if explicitly mentioned with `all` or `*` or something similar?

packages/pam/agent/accounts.go
```diff
@@ -0,0 +1,251 @@
...
+
+	if _, supported := supportedAccountTypes[account.AccountType]; !supported {
+		return fmt.Sprintf("%s accounts are not supported by the agent flow yet", account.AccountType)
+	}
+
+	if !account.CanLaunch {
+		return "you don't have permission to launch sessions for this account"
+	}
+
+	// An approval gate is deliberately not a reason to skip. RequestApprovals raises a request for
```
saifsmailbox98: IGNORE COMMENT PLACEMENT

we should not be getting an account a port if `--no-approval-request` is set, also the instruction tell the agent to keep retrying.

packages/pam/agent/instructions.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// RenderInstructions produces the markdown document handed to the agent.
+func RenderInstructions(accounts []LiveAccount) string {
+	var out strings.Builder
+
+	out.WriteString("# Infisical PAM: live privileged access\n\n")
+
+	fmt.Fprintf(&out, "%s available through local proxies on this machine. Connect to\n", pluralizeAccounts(len(accounts)))
+	out.WriteString("127.0.0.1 on the ports below. Any username and password will work: the real\n")
```
saifsmailbox98: only an empty password works for mysql

packages/pam/agent/instructions.go
```diff
@@ -0,0 +1,199 @@
...
+		return fmt.Sprintf("mysql://%s@127.0.0.1:%d", placeholderUser, port),
+			fmt.Sprintf("mysql -h 127.0.0.1 -P %d -u %s", port, placeholderUser)
+	case "mssql":
+		return fmt.Sprintf("sqlserver://%s@127.0.0.1:%d", placeholderUser, port),
+			fmt.Sprintf("sqlcmd -S 127.0.0.1,%d -U %s", port, placeholderUser)
+	case "mongodb":
+		return fmt.Sprintf("mongodb://127.0.0.1:%d", port),
+			fmt.Sprintf("mongosh --host 127.0.0.1 --port %d", port)
+	case "oracledb":
+		return fmt.Sprintf("%s@127.0.0.1:%d", placeholderUser, port),
```
saifsmailbox98: service name for oracle db is missing

packages/pam/agent/run.go
```diff
@@ -0,0 +1,528 @@
...
+	cwd, err := os.Getwd()
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve the working directory: %w", err)
+	}
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve the home directory: %w", err)
+	}
+
+	spec := sandbox.Spec{
```
saifsmailbox98, line 282: this spec never sets AllowTrustd, so on mac os trustd is denied and any go tool the agent runs (e.g., gh) fails TLS

saifsmailbox98, line 282: did you check if the we can allow the specific claude keychain item?

packages/pam/agent/accounts.go
```diff
@@ -0,0 +1,251 @@
...
+	byPath := make(map[string]api.PAMAccessibleAccount, len(accounts))
+	for _, account := range accounts {
+		byPath[strings.ToLower(accountPath(account))] = account
+	}
+
+	var resolved []ResolvedAccount
+	var problems []string
+	anyNotFound := false
+
+	for _, requested := range opts.Accounts {
```
saifsmailbox98, line 120: can we dedup `--account` if same account is used multiple times, instead of giving multiple ports and sessions to it? or just throw an error immediately?

packages/pam/agent/adapters.go
```diff
@@ -0,0 +1,253 @@
...
+		// Strip a Windows extension so "claude.exe" still matches.
+		name = strings.TrimSuffix(name, filepath.Ext(name))
+	}
+
+	for _, adapter := range adapters {
+		if strings.EqualFold(adapter.Name, name) {
+			return adapter
+		}
+	}
+	return fallbackAdapter
```
saifsmailbox98: if someone assumes that we support an agent let's say `opencode`, they'll try `--agent opencode` and it will silently use generic instead of throwing an error. let's throw an error if the input is invalid.

packages/pam/agent/run.go
```diff
@@ -0,0 +1,528 @@
...
+		path = filepath.Join(dir, "access.log")
+	}
+
+	file, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0o600)
+	if err != nil {
+		return "", fmt.Errorf("failed to open log file %s: %w", path, err)
+	}
+	s.cleanups = append(s.cleanups, func() { file.Close() })
+
+	log.Logger = zerolog.New(file).With().Timestamp().Logger()
```
saifsmailbox98, line 497: log.Logger is reassigned after proxies are already logging. Set up the logger before startProxies

packages/pam/agent/run.go
```diff
@@ -0,0 +1,528 @@
...
+}
+
+// redirectLogs sends proxy logging to a file so it cannot corrupt the agent's terminal UI.
+func (s *runSession) redirectLogs(path string) (string, error) {
+	if path == "" {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return "", fmt.Errorf("failed to resolve home directory: %w", err)
+		}
+		dir := filepath.Join(home, ".infisical", "pam-agentic")
```
saifsmailbox98: should we log without the user asking for it?

packages/pam/local/agent-transport.go
```diff
@@ -0,0 +1,44 @@
...
+//
+// Nothing in the interactive flow calls anything below.
+
+// NewAgentTransport builds the transport half of a proxy whose PAM session is created on demand.
+//
+// resourceType is what gets compared against the gateway's advertised capabilities, and provider
+// resolves the session behind each dial. ctx belongs to the caller, which owns its own shutdown,
+// because an agent's proxy ending must leave the process running.
+func NewAgentTransport(ctx context.Context, httpClient *resty.Client, resourceType string, provider SessionProvider) *BaseProxyServer {
+	return &BaseProxyServer{
```
saifsmailbox98, line 30: either set cancel from `context.WithCancel` or guard with if `b.cancel != nil`

packages/pam/agent/accounts.go
```diff
@@ -0,0 +1,251 @@
...
+	if len(paths) > limit {
+		return fmt.Sprintf("Accounts available to you:\n  - %s\n  ...and %d more",
+			strings.Join(paths[:limit], "\n  - "), len(paths)-limit)
+	}
+	return fmt.Sprintf("Accounts available to you:\n  - %s", strings.Join(paths, "\n  - "))
+}
+
+// listAllAccessibleAccounts pages through the accessible-accounts endpoint.
+func listAllAccessibleAccounts(httpClient *resty.Client) ([]api.PAMAccessibleAccount, error) {
+	const pageSize = 100
```
saifsmailbox98, line 286: what if the backend starts returning a different number of items?

packages/pam/agent/accounts.go
```diff
@@ -0,0 +1,251 @@
...
+	sort.Slice(skipped, func(i, j int) bool { return skipped[i].Path < skipped[j].Path })
+
+	if len(resolved) == 0 {
+		return nil, nil, errors.New(noUsableAccountsMessage(accounts, skipped))
+	}
+	return resolved, skipped, nil
+}
+
+// unusableReason explains why an account cannot be proxied, or returns "" if it can be.
+func unusableReason(account api.PAMAccessibleAccount, opts Options) string {
```
saifsmailbox98, line 180: for machine identities, MFA accounts fail cleanly but still get a port and fail on connect

packages/pam/agent/run.go
```diff
@@ -0,0 +1,528 @@
...
+// Only paths that already exist are returned, since bwrap --bind fails on a missing source.
+func agentStateWritePaths(home string) []string {
+	if home == "" {
+		return nil
+	}
+	candidates := []string{
+		filepath.Join(home, ".claude"),      // Claude Code state (sessions, history, cache)
+		filepath.Join(home, ".claude.json"), // Claude Code top-level config
+		filepath.Join(home, ".codex"),       // Codex state
+		filepath.Join(home, ".gemini"),      // Gemini CLI state
```
saifsmailbox98: can we put hostRuntimeDir and agentStateWritePaths as shared helpers in packages/sandbox? so that both agent proxy and and this can use it?

packages/pam/agent/accounts.go
```diff
@@ -0,0 +1,251 @@
...
+var supportedAccountTypes = map[string]string{
+	"postgres":   "PostgreSQL",
+	"mysql":      "MySQL",
+	"mssql":      "SQL Server",
+	"oracledb":   "Oracle",
+	"mongodb":    "MongoDB",
+	"redis":      "Redis",
+	"ssh":        "SSH",
+	"kubernetes": "Kubernetes",
+	"windows":    "Windows RDP",
```
saifsmailbox98: can an agent use an RDP session?

saifsmailbox98 (changes requested):

packages/cmd/pam.go
sheensantoscapadngan: I think the current behavior is the easiest one to onboard with... anyway, it doesn't start actual sessions until the agent makes a request relevant to a specific PAM account.. so I don't see any issues with starting the listeners ahead of time

saifsmailbox98 (approved):

PR #354, opened 2026-08-05
feat: KMIP server certificate auto-renewal token refresh

bernie-g:
# Description 📣

Bumps infisical-kmip to the version that auto-renews KMIP server TLS certificates (Infisical/infisical-kmip#14) and wires the token refresh hook so AWS-enrolled servers re-authenticate via STS when their access token is rejected. Tested E2E locally against a dev stack with a 3m-TTL certificate.

## Type ✨

- [x] New feature

### DIFF
--- a/packages/cmd/kmip.go
+++ b/packages/cmd/kmip.go
@@ -140,10 +140,6 @@ func startKmipServer(cmd *cobra.Command, args []string) {
 			if storedToken, _ := localkmip.LoadStoredAccessToken(serverName); storedToken != "" {
 				log.Info().Msg("Using stored KMIP server access token")
 				serverConfig.AccessToken = storedToken
-				// Keyed off the recorded enroll method rather than the server ID: LoadStoredServerID
-				// falls back to INFISICAL_KMIP_SERVER_ID, so a token-enrolled server with that
-				// variable set would be wired for STS refresh it cannot perform, and a rejected
-				// token would surface as an AWS failure instead of "re-enroll this server".
 				if serverID, canRefresh := localkmip.ResolveAwsRefreshServerID(serverName); canRefresh {
 					serverConfig.RefreshAccessToken = newAwsRefreshAccessTokenFunc(serverName, serverID)
 				}
--- a/packages/cmd/kmip.go
+++ b/packages/cmd/kmip.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
+	"time"
 
 	"github.com/Infisical/infisical-merge/packages/api"
 	"github.com/Infisical/infisical-merge/packages/config"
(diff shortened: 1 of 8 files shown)


packages/cmd/kmip.go
```diff
@@ -183,7 +186,26 @@ func enrollKmipServer(cmd *cobra.Command, enrollMethod, serverName string) strin
 		}
 
 		log.Info().Msgf("KMIP server authenticated via AWS Auth. State saved to %s", localkmip.GetConfPathDisplay(serverName))
-		return accessToken
+
+		refreshAccessToken := func() (string, error) {
+			refreshClient, err := util.GetRestyClientWithCustomHeaders()
```
carlosmonastyrski: This client has no time limit, so if the API never replies the refresh hangs and renewal is stuck behind it.

packages/cmd/kmip.go
carlosmonastyrski, line 142: Restarting an AWS server with no flags takes this path, which never sets the refresh function, so it can't recover on its own.

carlosmonastyrski (changes requested):

packages/cmd/kmip.go
bernie-g: Fixed

packages/cmd/kmip.go
bernie-g: Fixed

packages/cmd/kmip.go
```diff
@@ -138,6 +140,11 @@ func startKmipServer(cmd *cobra.Command, args []string) {
 			if storedToken, _ := localkmip.LoadStoredAccessToken(serverName); storedToken != "" {
 				log.Info().Msg("Using stored KMIP server access token")
 				serverConfig.AccessToken = storedToken
+				// Only AWS enrollment persists a server ID, so its presence means this server can
+				// re-authenticate via STS when the stored token is rejected mid-run.
+				if storedServerID, _ := localkmip.LoadStoredServerID(serverName); storedServerID != "" {
```
carlosmonastyrski: LoadStoredServerID reads the INFISICAL_KMIP_SERVER_ID env var before the config file, so a token-enrolled server can land here, fail AWS, and get “refreshing the token also failed” instead of the useful “re-enroll this server”. Check the saved enroll method instead.

packages/cmd/kmip.go
```diff
@@ -138,6 +140,13 @@ func startKmipServer(cmd *cobra.Command, args []string) {
 			if storedToken, _ := localkmip.LoadStoredAccessToken(serverName); storedToken != "" {
 				log.Info().Msg("Using stored KMIP server access token")
 				serverConfig.AccessToken = storedToken
+				// Keyed off the recorded enroll method rather than the server ID: LoadStoredServerID
+				// falls back to INFISICAL_KMIP_SERVER_ID, so a token-enrolled server with that
+				// variable set would be wired for STS refresh it cannot perform, and a rejected
```
carlosmonastyrski: LoadStoredServerID reads the INFISICAL_KMIP_SERVER_ID env var before the config file, so a token-enrolled server can land here, fail AWS, and get “refreshing the token also failed” instead of the useful “re-enroll this server”. Check the saved enroll method instead.

bernie-g: Fixed

carlosmonastyrski (dismissed):

carlosmonastyrski (approved):

PR #361, opened 2026-08-12
feat(agent-proxy): add policy-mode agent proxy for agent and user policy intersection

saifsmailbox98:
# Description 📣

Adds the runtime for [agent policies](https://github.com/Infisical/infisical/pull/7646): a long-standing agent proxy that brokers credentials on the intersection of an agent's policies and a user's.

`infisical agent-proxy start --token=<enrollment-token>` enrols against an agent proxy registered under Networking, swaps the one-time token for an access token (saved to `~/.infisical/agent-proxy/access-token` so a restart doesn't need a new one), signs a MITM intermediate off the org's agent proxy CA, and serves an HTTP forward proxy.

An agent points `HTTP_PROXY` at it with its session token as the proxy credential. Per request, the proxy resolves that session against Infisical and allows the request only when it matches at least one rule on an agent policy **and** at least one on a user policy. Neither side is a subset of the other, so this is a per-request check and not set arithmetic: a user rule of `GET` narrows an agent rule of `Any` on the same host. The matching agent policy is the most specific one, and that is what decides whose credentials get injected.

- `policy_match.go` — host patterns with scheme and HTTP method, and the intersection evaluator. A rule naming `https` refuses a plaintext request so a credential can't leave in the clear. `CONNECT` only carries host and port, so the tunnel opens on a host match and the real decision happens inside it once the method and path are known
- `policy_resolver.go` — per-session cache with a bounded size, refresh on a poll, and batched activity reporting. A hard auth failure drops the session and fails closed, so revocation needs no invalidation call
- `policy_server.go` — the proxy itself, reusing this package's existing `ca.go` and `rewrite.go` rather than forking them
- Blocked requests get a uniform 403 that doesn't say which side of the intersection refused

Companion PR: Infisical/infisical#7646

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Unit tests cover the evaluator, the pattern matcher and the proxy-auth parsing (`policy_match_test.go`), including the case the whole model rests on: agent allows `Any`, user allows `GET`, so the `POST` is refused and the `GET` is brokered.

Driven end to end against a local stack with one agent policy (Slack, any method) and one user policy (Slack, GET only):

```sh
GET  api.slack.com/api/auth.test        -> brokered, 200
POST api.slack.com/api/chat.postMessage -> 403, "no user policy allows this request"
CONNECT api.github.com                  -> 403, "no policy covers this host"
GET  registry.npmjs.org/express         -> passthrough, no credential (allowlist)
```

Substitution verified on the wire rather than inferred: with no credential Slack answers `not_authed`, and when the agent sends the placeholder it answers `invalid_auth`, so the upstream received a token the agent never held.

---

### DIFF
--- a/packages/agentproxy/policy_match.go
+++ b/packages/agentproxy/policy_match.go
@@ -1,0 +1,212 @@
+package agentproxy
+
+import (
+	"strings"
+)
+
+// policyPattern is a rule's host pattern plus the methods it covers. It is deliberately separate from
+// hostPattern (used by proxied services), which carries no scheme and no method.
+type policyPattern struct {
+	scheme  string
+	host    string
+	port    string
+	path    string
+	methods []string
+}
+
+// parsePolicyPattern accepts [scheme://]host[:port][/path*]. An unset scheme matches either scheme; a
+// set one is enforced, so a rule naming https never lets a plaintext request carry the credential.
+func parsePolicyPattern(raw string, methods []string) policyPattern {
+	p := policyPattern{methods: normalizeMethods(methods)}
+
+	part := strings.TrimSpace(raw)
+	if idx := strings.Index(part, "://"); idx != -1 {
+		p.scheme = strings.ToLower(part[:idx])
+		part = part[idx+3:]
+	}
+
+	if idx := strings.Index(part, "/"); idx != -1 {
+		p.path = part[idx:]
+		part = part[:idx]
+	}
+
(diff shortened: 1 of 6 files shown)


saifsmailbox98:
lisbon

PR #370, opened 2026-08-18
feat(gateway): load balancing

bernie-g:
# Description 📣

The gateway now keeps a count of how many connections it is currently handling and reports that number to Infisical every 10 seconds. This lets gateway pools send new work to whichever gateway is least busy, including work the platform never sees directly such as PAM sessions opened from the CLI.

Pairs with the platform change in Infisical/infisical#7703. Older gateways simply don't report, and the platform falls back to its own view for those.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Ran two gateways in a pool and watched the reported count while holding connections open. It rises with each connection, holds while they are open, and returns to zero when they close. Verified this includes a PAM session opened with `infisical pam access`, which the platform hands certificates to and never proxies itself.

```sh
# open a PAM session against an account on a gateway pool
infisical pam access df/my-account --duration 10m

# the gateway handling it reports its connection count
redis-cli get "gateway-reported-load:<gatewayId>"
# 2:1787065046473   <- two open connections
```

---


packages/gateway-v2/gateway.go
```diff
@@ -384,6 +393,36 @@ func (g *Gateway) reapIdleSessions() {
...
+
+		var last int64 = -1
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				count := g.activeChannels.Load()
+				// Still republish an unchanged count so the platform can tell a quiet gateway from
+				// one that has stopped reporting.
```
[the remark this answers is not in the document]
bernie-g: Fixed

packages/api/api.go
```diff
@@ -846,6 +848,25 @@ func CallGatewayHeartBeatV2(httpClient *resty.Client, request GatewayHeartbeatRe
...
+	if err != nil {
+		return NewGenericRequestError(operationCallGatewayLoadReportV2, err)
+	}
+
+	if response.IsError() {
+		return NewAPIErrorWithResponse(operationCallGatewayLoadReportV2, response, nil)
+	}
+
+	return nil
+}
```
sheensantoscapadngan: hmm... can't we track this from the platform side so gateway doesn't have to reach out to Infisical so often?

packages/gateway-v2/gateway.go
```diff
@@ -884,6 +936,9 @@ func (g *Gateway) handleIncomingChannel(newChannel ssh.NewChannel) {
 	}
 	defer channel.Close()
 
+	g.activeChannels.Add(1)
```
lb-vn: Claude:

This counter only goes down via the defer, so a leaked goroutine inflates it permanently and the gateway stops getting work. Could we reset it to 0 on each new relay connection?

packages/gateway-v2/gateway.go
```diff
@@ -384,6 +399,42 @@ func (g *Gateway) reapIdleSessions() {
...
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				count := g.activeChannels.Load()
+				// Still republish an unchanged count so the platform can tell a quiet gateway from
+				// one that has stopped reporting.
+				if err := g.sendLoadReport(ctx, count); err != nil {
+					log.Debug().Msgf("Load report failed: %v", err)
```
lb-vn: An older self hosted platform has no metrics endpoint, so this retries every 10s forever. Maybe we should have some back-off mechanism that stops the report after X failures?

packages/gateway-v2/gateway.go
```diff
@@ -384,6 +399,42 @@ func (g *Gateway) reapIdleSessions() {
...
+// the platform falls back to its own view when a gateway stops reporting.
+func (g *Gateway) sendLoadReport(ctx context.Context, count int64) error {
+	reqCtx, cancel := context.WithTimeout(ctx, loadReportTimeout)
+	defer cancel()
+	return api.CallGatewayLoadReportV2(reqCtx, g.httpClient, api.GatewayLoadReportRequest{ActiveChannels: count})
+}
+
+func (g *Gateway) reportLoad(ctx context.Context) {
+	go func() {
+		ticker := time.NewTicker(loadReportInterval)
```
x032205: the first report doesn't land until 10s after the relay connects, and while it's missing isPoolComparable turns load balancing off for the whole pool, not just this member. that means every gateway restart costs the entire pool 10s of random selection. we should send one report immediately before entering this loop.

packages/gateway-v2/gateway.go
```diff
@@ -384,6 +399,42 @@ func (g *Gateway) reapIdleSessions() {
 	}
 }
 
+// reportLoad publishes the gateway's active channel count so the platform can route new work to the
```
lb-vn: The comment mentions "reportLoad" but this is the "sendMetricsReport" function

packages/gateway-v2/gateway.go
```diff
@@ -384,6 +399,42 @@ func (g *Gateway) reapIdleSessions() {
...
+// least loaded member of a pool. Failures are not surfaced: a missed report only costs accuracy, and
+// the platform falls back to its own view when a gateway stops reporting.
+func (g *Gateway) sendLoadReport(ctx context.Context, count int64) error {
+	reqCtx, cancel := context.WithTimeout(ctx, loadReportTimeout)
+	defer cancel()
+	return api.CallGatewayLoadReportV2(reqCtx, g.httpClient, api.GatewayLoadReportRequest{ActiveChannels: count})
+}
+
+func (g *Gateway) reportLoad(ctx context.Context) {
```
x032205: this doesn't report, it starts the reporter and returns immediately. we should call it startLoadReporting so it matches startHeartbeatOnce and the call site reads as what it does.

packages/api/api.go
bernie-g: Resolved on slack

packages/gateway-v2/gateway.go
lb-vn: Also this comment may be for "reportMetrics" instead of "sendMetricsReport"? Not 100% sure

lb-vn (changes requested):

packages/gateway-v2/gateway.go
lb-vn: Also maybe we should rename "reportMetrics" to "startMetricsReport" to be more clear

packages/gateway-v2/gateway.go
bernie-g: Fixed

packages/gateway-v2/gateway.go
bernie-g: Added a backoff

packages/gateway-v2/gateway.go
bernie-g: Fixed

packages/gateway-v2/gateway.go
```diff
@@ -884,6 +967,10 @@ func (g *Gateway) handleIncomingChannel(newChannel ssh.NewChannel) {
 	}
 	defer channel.Close()
 
+	generation := g.channelGeneration.Load()
```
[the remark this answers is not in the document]
bernie-g: Fixed

lb-vn (approved):

PR #372, opened 2026-08-20
Add named login profiles and directory-scoped tenant selection

scott-ray-wilson:
## Problem

Switching organizations means logging out and back in. The CLI keeps one session per account, keyed by email in the keyring, and a session token is scoped to a single organization, so a second login destroys the first. Working across tenants in parallel means exporting tokens into env vars or `.env` files. Community ask: Infisical/infisical#2191.

## Model

A **profile** is one login: an account on one instance, plus the organization it uses by default. Each profile gets its own keyring entry, so sessions coexist.

The **organization is a field of the profile**, not part of its identity. `--org` / `INFISICAL_ORG` retarget a single command by name, slug or id, backed by a per-organization token cache, so the switch costs one exchange and nothing after. `profile set-org` (aka `org switch`) changes the profile's default.

Selection per invocation: `--profile` > `INFISICAL_PROFILE` > bound directory > machine default. Those last three each get a verb, so all are discoverable from `profile --help`:

| Scope | Command | Undo |
|---|---|---|
| This machine | `profile use <name>` | reassign |
| This terminal | `profile pin <name>` (via `eval`) | `profile unpin` |
| This directory tree | `profile bind [name] [path]` | `profile unbind` |

An explicit override beats a bound directory and says so, so a binding that did not apply is explained rather than silently ignored.

Profiles are named `<email>--<org-slug>` (`scott@example.com--acme-x4k2`), so a name says which tenant it is without listing anything. `profile rename` changes it.

## Commands

```
profile list | current | new | use | pin | unpin | bind | unbind | set-org | rename | delete
org list | switch
login [--save-as <name>] [--profile <name>]
logout [--all] [--local-only]
```

`login --save-as <name>` stores the login under a name you choose, creating that profile or replacing its session. `login --profile <name>` signs back in to a profile that already exists, keeping its account, instance and organization, so `--organization-id` is not needed again; naming a profile that does not exist fails and points at `--save-as`.

## Rationale for the less obvious calls

- **Org as a field, not identity.** Reviewers of the first cut said `org switch` and profiles felt like unrelated features. They were: org lived inside the profile's identity, so changing it meant mutating or forking a profile. Making it a field is the kubectl context/namespace split.
- **Targeted operations never move the machine default.** `login --save-as`, `login --profile`, `profile new`, `org switch` from a pinned shell all leave it alone. The default is what every unpinned terminal resolves to, so moving it reaches across terminals. `--use` opts in explicitly.
- **`--org` resolves locally first.** Profiles record the organization slug, and organizations reached before are indexed on the profile, each matched by id, slug or name. A repeated `--org` therefore makes no API calls at all; only the first use of a new organization lists and exchanges. The trade-off is that a name matching your own organization short-circuits without asking the server, so a same-named organization you have never used is no longer detected. Ids and slugs are unique, so they are unaffected.
- **Organization tokens are separate keyring entries.** Keeping them inside the profile's own entry would grow it past the platform limit (roughly 4 KB on macOS) after a few organizations, and the write would start failing silently. Each cached session is now `org-session:<profile>:<orgId>`; the profile stores only id, name and slug, which is what a selector matches on.
- **Sub-organizations.** A session scoped to one carries the root organization in `organizationId` and the sub-organization in `subOrganizationId`, and acts in the latter. Profiles record both and expose the scoped one, which is what `init` filters projects by, what `org list` marks as current, and what `--org` compares against.
- **Sessions still expire at `JWT_AUTH_LIFETIME`; renewal stays unimplemented.** Unchanged from `main`, where the refresh path is commented out. The server rotates the refresh token on every refresh and treats a stale one as theft by revoking the session, with a 10s grace window, which several CLI processes sharing one vault entry cannot coordinate. The token is also no longer persisted, since nothing read it.
- **Logout respects shared sessions.** The server keys sessions by user, IP and user agent, so profiles for one account on one machine share a session. Logout leaves a session another profile still uses intact.
- **`init` stops asking for an organization.** It takes the profile's. The old prompt also silently re-scoped your session as a side effect.
- **MFA.** `profile new` and `profile set-org` perform their own exchange and prompt. `--org` on an ordinary command runs inside credential resolution and cannot prompt, so it fails with a message naming the command that can.
- **The "Using profile" notice lives in the session loader.** It prints the first time a command actually loads a login session, so `scan`, `agent` or `gateway` stay quiet even in a pinned terminal.

## Compatibility

Migration is lazy and needs no re-login: legacy `loggedInUserEmail` / `loggedInUsers` become profiles named after the account email, which is already the keyring key, so existing sessions keep working.

The legacy fields are only published for a profile named after its email. Migrated profiles still are, so a `main` binary keeps working against them. A profile created by this branch is named `<email>--<org-slug>`, so a `main` binary will ask for a fresh login rather than load some other profile's token against this profile's instance.

**Behavior changes for release notes:**

- `init` no longer prompts for an organization (`--org` or `profile set-org` to cross).
- `init` now **fails** instead of warning when `--org` would link a project that later commands, using the profile's own organization, could not find.
- `login --profile <name>` no longer creates a profile; use `login --save-as <name>`.
- `reset` now revokes sessions server-side instead of only deleting local files.
- `vault set` clears the profile list along with the login it already cleared.
- `login status` reports the profile, how it was selected, and the organization name; `--json` gains `profile`, `profileSource` and `organizationName`.

## Verification

Automated script covering 20 behaviors (login and migration, targeted-login isolation, org-as-a-dimension including token-repointing, `profile new` with `--use`/`--pin`, the full precedence chain, pin's refusal to report false success, logout revocation including the shared-session case). Runs in a throwaway `HOME`, needs an account with two orgs. Script is attached in a comment below.

On top of that, a **93-check matrix run against a local instance**, one section per review comment, seeded with two root organizations, a third whose slug differs from its name, and a **sub-organization with its own project**. All 93 pass. It covers:

- `init` on a sub-organization session listing and linking that organization's project, and a root-organization session still seeing only its own.
- `--org` API call volume, counted from the access log: 0 calls for your own organization by id, slug or name including first use; 1 listing + 1 exchange for a new organization, then 0.
- Cached organization sessions landing in their own keyring entries, with the profile's entry staying near 1 KB across three cached organizations.
- A real email-MFA login driven end to end, plus the `--org` refusal that points at `profile set-org`.
- Pin and unpin behavior on a real TTY, profile rename moving sessions and bindings, and name length limits.

Done manually against a local instance as well: sub-org resolution by name and slug, single-profile output diffed byte-for-byte against a `main` build, two terminals pinned to different profiles, and per-profile instance routing.

Worth a reviewer's own pass:

- [ ] Two terminals pinned to different profiles, then change the default in a third; the pinned two must not move
- [ ] Multi-instance: profiles on cloud and self-hosted, each command reaching its own host with no `--domain`
- [ ] Browser login (only the interactive flow shows the org picker in the terminal)
- [ ] Upgrade from a real pre-profiles config with several accounts

## Found while testing, pre-existing on `main`, not fixed here

- A wrong MFA code never prints "You have N attempts left". The CLI matches `context.code == "mfa_invalid"`, but the backend now returns a plain 500 with no code, so it silently re-prompts.
- `user get token` overwrites the session error with the result of reading its own `--plain` flag, so a failed session surfaces as "invalid token format".

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/api/api.go
+++ b/packages/api/api.go
@@ -77,6 +77,7 @@ const (
 	operationCallGetCertificateBundle              = "CallGetCertificateBundle"
 	operationCallRenewCertificate                  = "CallRenewCertificate"
 	operationCallGetCertificateRequest             = "CallGetCertificateRequest"
+	operationCallRevokeUserSession                 = "CallRevokeUserSession"
 )
 
 var ErrNotFound = errors.New("resource not found")
@@ -158,6 +159,25 @@ func CallLoginV3(httpClient *resty.Client, request GetLoginV3Request) (GetLoginV
 	}
 
 	return loginV3Response, nil
+}
+
+// CallRevokeUserSession revokes a single server-side login session by its id
+// (the tokenVersionId claim carried in every session JWT).
+func CallRevokeUserSession(httpClient *resty.Client, sessionID string) error {
+	response, err := httpClient.
+		R().
+		SetHeader("User-Agent", USER_AGENT).
+		Delete(fmt.Sprintf("%v/v2/users/me/sessions/%v", config.INFISICAL_URL, url.PathEscape(sessionID)))
+
+	if err != nil {
+		return NewGenericRequestError(operationCallRevokeUserSession, err)
+	}
+
+	if response.IsError() {
+		return NewAPIErrorWithResponse(operationCallRevokeUserSession, response, nil)
+	}
+
+	return nil
 }
 
 func CallVerifyMfaToken(httpClient *resty.Client, request VerifyMfaTokenRequest) (*VerifyMfaTokenResponse, *VerifyMfaTokenErrorResponse, error) {
--- a/packages/cmd/init.go
+++ b/packages/cmd/init.go
@@ -6,6 +6,7 @@ package cmd
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 
 	"github.com/Infisical/infisical-merge/packages/api"
 	"github.com/Infisical/infisical-merge/packages/config"
@@ -57,64 +58,66 @@ var initCmd = &cobra.Command{
 		}
 		httpClient.SetAuthToken(userCreds.UserCredentials.JTWToken)
 
-		selectedOrgID, selectedSubOrgName, err := pickOrganization(httpClient, "Which Infisical organization would you like to select a project from?", userCreds.UserCredentials.Email)
-		if err != nil {
-			util.HandleError(err, "Unable to select organization")
-		}
+		// The profile already carries an organization (and --org can retarget it
+		// for this command), so don't ask again. Only fall back to the picker
+		// when the profile has no organization recorded, which happens for
+		// sessions migrated from a CLI that predates profiles.
+		selectedOrgID := userCreds.OrganizationID
+		var selectedSubOrgName *string
 
-		tokenResponse, err := api.CallSelectOrganization(httpClient, api.SelectOrganizationRequest{OrganizationId: selectedOrgID})
-		if tokenResponse.MfaEnabled {
-			i := 1
-			for i < 6 {
-				mfaVerifyCode := askForMFACode(tokenResponse.MfaMethod)
-
-				httpClient, err := util.GetRestyClientWithCustomHeaders()
-				if err != nil {
-					util.HandleError(err, "Unable to get resty client with custom headers")
-				}
-				httpClient.SetAuthToken(tokenResponse.Token)
-				verifyMFAresponse, mfaErrorResponse, requestError := api.CallVerifyMfaToken(httpClient, api.VerifyMfaTokenRequest{
-					Email:     userCreds.UserCredentials.Email,
-					MFAToken:  mfaVerifyCode,
-					MFAMethod: tokenResponse.MfaMethod,
-				})
-				if requestError != nil {
-					util.HandleError(err)
-					break
-				} else if mfaErrorResponse != nil {
-					if mfaErrorResponse.Context.Code == "mfa_invalid" {
-						msg := fmt.Sprintf("Incorrect, verification code. You have %v attempts left", 5-i)
-						util.PrintlnStderr(msg)
-						if i == 5 {
-							util.PrintErrorMessageAndExit("No tries left, please try again in a bit")
-							break
-						}
-					}
-
-					if mfaErrorResponse.Context.Code == "mfa_expired" {
-						util.PrintErrorMessageAndExit("Your 2FA verification code has expired, please try logging in again")
-						break
-					}
-					i++
-				} else {
-					httpClient.SetAuthToken(verifyMFAresponse.Token)
-					tokenResponse, err = api.CallSelectOrganization(httpClient, api.SelectOrganizationRequest{OrganizationId: selectedOrgID})
-					break
-				}
+		if selectedOrgID == "" {
+			pickedOrgID, pickedSubOrgName, err := pickOrganization(httpClient, "Which Infisical organization would you like to select a project from?", userCreds.UserCredentials.Email)
+			if err != nil {
+				util.HandleError(err, "Unable to select organization")
 			}
-		}
+			selectedSubOrgName = pickedSubOrgName
 
-		if err != nil {
-			util.HandleError(err, "Unable to select organization")
-		}
+			newSessionToken, err := selectOrganizationToken(userCreds.UserCredentials.JTWToken, userCreds.UserCredentials.Email, pickedOrgID)
+			if err != nil {
+				util.HandleError(err, "Unable to select organization")
+			}
 
-		// set the config jwt token to the new token
-		userCreds.UserCredentials.JTWToken = tokenResponse.Token
-		err = util.StoreUserCredsInKeyRing(&userCreds.UserCredentials)
-		httpClient.SetAuthToken(tokenResponse.Token)
+			// The session token is now scoped to the selected organization; record
+			// it on the profile this invocation resolved to so later commands in
+			// this project don't have to ask again.
+			userCreds.UserCredentials.JTWToken = newSessionToken
+			orgID, subOrgID := util.ParseTokenOrgClaims(newSessionToken)
+			if orgID == "" {
+				orgID = pickedOrgID
+			}
+			selectedOrgID = orgID
 
-		if err != nil {
-			util.HandleError(err, "Unable to store your user credentials")
+			updatedProfile := userCreds.Profile
+			updatedProfile.OrganizationID = orgID
+			updatedProfile.SubOrganizationID = subOrgID
+			updatedProfile.OrganizationName = util.OrgDisplayName(newSessionToken, orgID, subOrgID)
+
+			// Only move the global default when this invocation was using it; a
+			// terminal pinned via env var, flag, or directory scope must not switch
+			// other terminals.
+			makeActive := userCreds.ProfileSource == util.ProfileSourceDefault
+			err = util.PersistLoginProfile(updatedProfile, &userCreds.UserCredentials, makeActive)
+			httpClient.SetAuthToken(newSessionToken)
+
+			if err != nil {
+				util.HandleError(err, "Unable to store your user credentials")
+			}
+		} else {
+			orgDisplay := userCreds.OrganizationName
+			if orgDisplay == "" {
+				orgDisplay = selectedOrgID
+			}
+			util.PrintlnStderr(fmt.Sprintf("Using organization %s from profile '%s'. Pass --org to pick a different one.", orgDisplay, userCreds.ProfileName))
+
+			// An --org override is per command, so a project linked under it would
+			// not resolve on later runs that use the profile's default.
+			if userCreds.OrganizationSource != util.OrgSourceProfileDefault && userCreds.Profile.OrganizationID != "" && userCreds.OrganizationID != userCreds.Profile.OrganizationID {
+				profileOrg := userCreds.Profile.OrganizationName
+				if profileOrg == "" {
+					profileOrg = userCreds.Profile.OrganizationID
+				}
+				util.PrintWarning(fmt.Sprintf("Profile '%s' defaults to organization %s, so later commands here will not find this project unless you pass --org again. Run [infisical profile set-org %s] to make it the default.", userCreds.ProfileName, profileOrg, orgDisplay))
+			}
 		}
 
 		workspaceResponse, err := api.CallGetAllWorkSpacesUserBelongsTo(httpClient)
@@ -140,9 +143,46 @@ var initCmd = &cobra.Command{
 			util.HandleError(err)
 		}
 
+		offerDirectoryProfileBinding(userCreds.ProfileName)
+
 		Telemetry.CaptureEvent("cli-command:init", posthog.NewProperties().Set("version", util.CLI_VERSION))
 
 	},
+}
+
+// offerDirectoryProfileBinding asks (only when multiple profiles exist)
+// whether this directory should always use the profile init just ran with, so
+// commands run here pick the right tenant without flags or env vars.
+func offerDirectoryProfileBinding(profileName string) {
+	configFile, err := util.GetMigratedConfigFile()
+	if err != nil || profileName == "" || len(configFile.Profiles) < 2 {
+		return
+	}
+
+	cwd, err := os.Getwd()
+	if err != nil {
+		return
+	}
+
+	if boundProfile, _, ok := util.FindGoverningDirectoryProfile(configFile, cwd); ok && boundProfile == profileName {
+		return
+	}
+
+	prompt := promptui.Select{
+		Label: fmt.Sprintf("Bind this directory to profile '%s'? Commands run here will then select it automatically. Select[Yes/No]", profileName),
+		Items: []string{"No", "Yes"},
+	}
+	_, result, err := prompt.Run()
+	if err != nil || result != "Yes" {
+		return
+	}
+
+	util.SetDirectoryProfile(&configFile, cwd, profileName)
+	if err := util.WriteConfigFile(&configFile); err != nil {
+		util.PrintWarning(fmt.Sprintf("Unable to save the directory profile binding [err=%s]", err))
+		return
+	}
+	util.PrintlnStderr(fmt.Sprintf("Directory %s now uses profile '%s'. Manage bindings with [infisical profile bind] and [infisical profile unbind].", cwd, profileName))
 }
 
 func init() {
--- a/packages/cmd/init.go
+++ b/packages/cmd/init.go
@@ -62,8 +62,15 @@ var initCmd = &cobra.Command{
 		// for this command), so don't ask again. Only fall back to the picker
 		// when the profile has no organization recorded, which happens for
 		// sessions migrated from a CLI that predates profiles.
+		//
+		// OrganizationID is the organization the session acts in. For a session
+		// scoped to a sub-organization that is the sub-organization, which is
+		// also where its projects are filed, not the root from the token.
 		selectedOrgID := userCreds.OrganizationID
 		var selectedSubOrgName *string
+		if parent, own := util.SplitOrgDisplayName(userCreds.OrganizationName); parent != "" {
+			selectedSubOrgName = &own
+		}
 
 		if selectedOrgID == "" {
 			pickedOrgID, pickedSubOrgName, err := pickOrganization(httpClient, "Which Infisical organization would you like to select a project from?", userCreds.UserCredentials.Email)
@@ -85,12 +92,14 @@ var initCmd = &cobra.Command{
 			if orgID == "" {
 				orgID = pickedOrgID
 			}
-			selectedOrgID = orgID
+			orgInfo := util.DescribeSessionOrg(newSessionToken, orgID, subOrgID)
 
 			updatedProfile := userCreds.Profile
 			updatedProfile.OrganizationID = orgID
 			updatedProfile.SubOrganizationID = subOrgID
-			updatedProfile.OrganizationName = util.OrgDisplayName(newSessionToken, orgID, subOrgID)
+			updatedProfile.OrganizationName = orgInfo.DisplayName()
+			updatedProfile.OrganizationSlug = orgInfo.Slug
+			selectedOrgID = updatedProfile.ScopedOrganizationID()
 
 			// Only move the global default when this invocation was using it; a
 			// terminal pinned via env var, flag, or directory scope must not switch
@@ -107,17 +116,23 @@ var initCmd = &cobra.Command{
 			if orgDisplay == "" {
 				orgDisplay = selectedOrgID
 			}
-			util.PrintlnStderr(fmt.Sprintf("Using organization %s from profile '%s'. Pass --org to pick a different one.", orgDisplay, userCreds.ProfileName))
 
-			// An --org override is per command, so a project linked under it would
-			// not resolve on later runs that use the profile's default.
-			if userCreds.OrganizationSource != util.OrgSourceProfileDefault && userCreds.Profile.OrganizationID != "" && userCreds.OrganizationID != userCreds.Profile.OrganizationID {
+			// An --org override is per command, so a project linked under it
+			// would not resolve on later runs that use the profile's default.
+			// That would surface later as an unrelated-looking "project not
+			// found", and a warning here can be silenced, so refuse and point
+			// at the two ways to make the organization stick.
+			if userCreds.OrganizationSource != util.OrgSourceProfileDefault && userCreds.Profile.OrganizationID != "" && userCreds.OrganizationID != userCreds.Profile.ScopedOrganizationID() {
 				profileOrg := userCreds.Profile.OrganizationName
 				if profileOrg == "" {
-					profileOrg = userCreds.Profile.OrganizationID
+					profileOrg = userCreds.Profile.ScopedOrganizationID()
 				}
-				util.PrintWarning(fmt.Sprintf("Profile '%s' defaults to organization %s, so later commands here will not find this project unless you pass --org again. Run [infisical profile set-org %s] to make it the default.", userCreds.ProfileName, profileOrg, orgDisplay))
+				util.PrintErrorMessageAndExit(
+					fmt.Sprintf("Profile '%s' defaults to organization %s, so a project linked here under %s (selected via %s) would not be found by later commands unless they also pass --org.", userCreds.ProfileName, profileOrg, orgDisplay, userCreds.OrganizationSource),
+					fmt.Sprintf("Make %s the profile's default with [infisical profile set-org %s], or keep both organizations by running [infisical profile new <name> --org %s] and then [infisical profile bind <name>] in this directory.", orgDisplay, orgDisplay, orgDisplay))
 			}
+
+			util.PrintlnStderr(fmt.Sprintf("Using organization %s from profile '%s'. Pass --org to pick a different one.", orgDisplay, userCreds.ProfileName))
 		}
 
 		workspaceResponse, err := api.CallGetAllWorkSpacesUserBelongsTo(httpClient)
--- a/packages/cmd/login.go
+++ b/packages/cmd/login.go
@@ -134,14 +134,17 @@ var loginCmd = &cobra.Command{
 			}
 
 			currentLoggedInUserDetails, err := util.GetCurrentLoggedInUserDetails(true)
-			// if the key can't be found or there is an error getting current credentials from key ring, allow them to override
-			if err != nil && (strings.Contains(err.Error(), "we couldn't find your logged in details")) {
+			// if the key can't be found, the selected profile doesn't exist yet, or
+			// there is an error getting current credentials from key ring, allow them to override
+			if err != nil && (errors.Is(err, util.ErrProfileNotFound) || errors.Is(err, util.ErrProfileDomainMismatch) || strings.Contains(err.Error(), "we couldn't find your logged in details")) {
 				log.Debug().Err(err)
 			} else if err != nil {
 				util.HandleError(err)
 			}
 
-			if currentLoggedInUserDetails.IsUserLoggedIn && !currentLoggedInUserDetails.LoginExpired && len(currentLoggedInUserDetails.UserCredentials.PrivateKey) != 0 {
+			// When a profile is explicitly targeted (flag or env var), the login is
+			// a deliberate write to that profile; skip the add/override menu.
+			if config.INFISICAL_PROFILE_OVERRIDE == "" && currentLoggedInUserDetails.IsUserLoggedIn && !currentLoggedInUserDetails.LoginExpired && len(currentLoggedInUserDetails.UserCredentials.PrivateKey) != 0 {
 				shouldOverride, err := userLoginMenu(currentLoggedInUserDetails.UserCredentials.Email)
 				if err != nil {
 					util.HandleError(err)
@@ -227,18 +230,46 @@ var loginCmd = &cobra.Command{
 				cliDefaultLogin(&userCredentialsToBeStored, email, password, organizationId)
 			}
 
-			err = util.StoreUserCredsInKeyRing(&userCredentialsToBeStored)
+			orgID, subOrgID := util.ParseTokenOrgClaims(userCredentialsToBeStored.JTWToken)
+			orgName := util.OrgDisplayName(userCredentialsToBeStored.JTWToken, orgID, subOrgID)
+
+			existingConfig, err := util.GetMigratedConfigFile()
+			if err != nil {
+				util.HandleError(err, "Unable to read the Infisical config file")
+			}
+
+			profileName := config.INFISICAL_PROFILE_OVERRIDE
+			if profileName == "" {
+				profileName = util.DeriveProfileName(existingConfig, userCredentialsToBeStored.Email, config.INFISICAL_URL, orgID, orgName)
+			} else if err := util.ValidateProfileName(profileName); err != nil {
+				util.HandleError(err)
+			}
+
+			if existingProfile, found := util.FindProfile(existingConfig, profileName); found && existingProfile.Email != userCredentialsToBeStored.Email {
+				util.PrintWarning(fmt.Sprintf("Profile '%s' previously stored the session for %s and now stores the session for %s.", profileName, existingProfile.Email, userCredentialsToBeStored.Email))
+			}
+
+			// An explicitly targeted login (--profile flag or INFISICAL_PROFILE) is a
+			// scoped write: it must not move the global default out from under other
+			// terminals that rely on it. This also keeps expired-session renewals
+			// (which re-exec login with --profile) from stealing the default.
+			// Untargeted logins keep the familiar "last login wins" behavior.
+			makeActive := config.INFISICAL_PROFILE_OVERRIDE == ""
+
+			err = util.PersistLoginProfile(models.Profile{
+				Name:              profileName,
+				Email:             userCredentialsToBeStored.Email,
+				Domain:            config.INFISICAL_URL,
+				OrganizationID:    orgID,
+				OrganizationName:  orgName,
+				SubOrganizationID: subOrgID,
+			}, &userCredentialsToBeStored, makeActive)
 			if err != nil {
 				log.Error().Msgf("Unable to store your credentials in system vault")
 				log.Error().Msgf("\nTo trouble shoot further, read https://infisical.com/docs/cli/faq")
 				log.Debug().Err(err)
 				//return here
 				util.HandleError(err)
-			}
-
-			err = util.WriteInitalConfig(&userCredentialsToBeStored)
-			if err != nil {
-				util.HandleError(err, "Unable to write write to Infisical Config file. Please try again")
 			}
 
 			// Identify the user in PostHog and alias the anonymous machine ID
@@ -266,6 +297,17 @@ var loginCmd = &cobra.Command{
 			time.Sleep(time.Second * 1)
 			boldWhite.Printf(">>>> Welcome to Infisical!")
 			boldWhite.Printf(" You are now logged in as %v <<<< \n", userCredentialsToBeStored.Email)
+
+			if profileName != userCredentialsToBeStored.Email {
+				orgDetail := ""
+				if orgName != "" {
+					orgDetail = fmt.Sprintf(" (org %s)", orgName)
+				}
+				util.PrintlnStderr(fmt.Sprintf("Session saved to profile '%s'%s. Select it with --profile %s or INFISICAL_PROFILE=%s.", profileName, orgDetail, profileName, profileName))
+			}
+			if configAfterLogin, err := util.GetConfigFile(); err == nil && configAfterLogin.ActiveProfile != "" && configAfterLogin.ActiveProfile != profileName {
+				util.PrintlnStderr(fmt.Sprintf("Your default profile remains '%s'; terminals using it are unaffected. Run [infisical profile use %s] to make '%s' the default.", configAfterLogin.ActiveProfile, profileName, profileName))
+			}
 
 			plainBold := color.New(color.Bold)
 
--- a/packages/cmd/login.go
+++ b/packages/cmd/login.go
@@ -48,10 +48,21 @@ const REPLACE_USER = "Override current logged in user"
 const EXIT_USER_MENU = "Exit"
 const QUIT_BROWSER_LOGIN = "q"
 
+// mfaMaxAttempts is how many verification codes a user may try before a
+// command gives up. Every MFA prompt in the CLI uses it, so they stay in step.
+const mfaMaxAttempts = 5
+
 // loginCmd represents the login command
 var loginCmd = &cobra.Command{
-	Use:                   "login",
-	Short:                 "Login into your Infisical account",
+	Use:   "login",
+	Short: "Login into your Infisical account",
+	Long: `Login into your Infisical account.
+
+Sessions are stored in login profiles. Without options the profile is named
+after the account and organization, as in scott@example.com--acme-x4k2. Use
+--save-as <name> to store the login under a name of your choice, creating that
+profile or replacing its session, and --profile <name> to sign back in to a
+profile that already exists.`,
 	DisableFlagsInUseLine: true,
 	PreRunE: func(cmd *cobra.Command, args []string) error {
 		// daniel: oidc-jwt is deprecated in favor of `jwt`. we backfill the `jwt` flag with the value of `oidc-jwt` if it's set.
@@ -127,13 +138,40 @@ var loginCmd = &cobra.Command{
 
 		// standalone user auth
 		if loginMethod == "user" {
-			isDirectUserLoginFlagsAndEnvsSet, err := validateDirectUserLoginFlagsAndEnvsSet(cmd, presetDomain)
+			existingConfig, err := util.GetMigratedConfigFile()
+			if err != nil {
+				util.HandleError(err, "Unable to read the Infisical config file")
+			}
+
+			saveAs, err := cmd.Flags().GetString("save-as")
+			if err != nil {
+				util.HandleError(err)
+			}
+			profileOverride, profileOverrideSource := util.GetProfileOverride()
+			target, err := resolveLoginTarget(saveAs, profileOverride, profileOverrideSource, existingConfig)
+			if err != nil {
+				util.HandleError(err)
+			}
+
+			// Signing back in to a profile takes its organization, so
+			// --organization-id is not required to drive the login
+			// non-interactively the way it is for a fresh one.
+			isDirectUserLoginFlagsAndEnvsSet, err := validateDirectUserLoginFlagsAndEnvsSet(cmd, presetDomain, target.reauth && target.profile.ScopedOrganizationID() != "")
 
 			if err != nil {
 				util.HandleError(err)
 			}
 
+			// Login creates or refreshes a session and never scopes it through
+			// --org, so resolve the existing session with the override suspended:
+			// applying it here could only fail (unknown organization, MFA) and
+			// abort the login.
+			if _, orgSource := util.GetOrgOverride(); orgSource == util.OrgSourceFlag && !silentMode {
+				util.PrintWarning("--org does not apply to login. Pick the organization in the prompt, or pass --organization-id.")
+			}
+			restoreOrgOverride := util.SuspendOrgOverride()
 			currentLoggedInUserDetails, err := util.GetCurrentLoggedInUserDetails(true)
+			restoreOrgOverride()
 			// if the key can't be found, the selected profile doesn't exist yet, or
 			// there is an error getting current credentials from key ring, allow them to override
 			if err != nil && (errors.Is(err, util.ErrProfileNotFound) || errors.Is(err, util.ErrProfileDomainMismatch) || strings.Contains(err.Error(), "we couldn't find your logged in details")) {
@@ -142,9 +180,10 @@ var loginCmd = &cobra.Command{
 				util.HandleError(err)
 			}
 
-			// When a profile is explicitly targeted (flag or env var), the login is
-			// a deliberate write to that profile; skip the add/override menu.
-			if config.INFISICAL_PROFILE_OVERRIDE == "" && currentLoggedInUserDetails.IsUserLoggedIn && !currentLoggedInUserDetails.LoginExpired && len(currentLoggedInUserDetails.UserCredentials.PrivateKey) != 0 {
+			// A login that names its profile (--save-as, or --profile for an
+			// existing one) is a deliberate write to that profile; skip the
+			// add/override menu.
+			if !target.explicit && currentLoggedInUserDetails.IsUserLoggedIn && !currentLoggedInUserDetails.LoginExpired && len(currentLoggedInUserDetails.UserCredentials.PrivateKey) != 0 {
 				shouldOverride, err := userLoginMenu(currentLoggedInUserDetails.UserCredentials.Email)
 				if err != nil {
 					util.HandleError(err)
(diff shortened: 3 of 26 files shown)


scott-ray-wilson:
Addressed the review findings and root-caused the CI failures in c0a430e:

**CI: Go CLI Tests hang (caused by this PR, now fixed).** The harness drives `login --interactive` then `init` through a pty and never checks login's exit code. On this branch, login authenticated successfully and then died locally: the derived profile name is the raw account email, and `ValidateProfileName` rejected characters like `+` that are legal in emails (the CI test account's email evidently contains one). With no session stored, `init`'s `RequireLogin` auto-spawned `login --silent`, whose "Select your hosting option" prompt the harness has no answer for, deadlocking until the 11m test timeout. Reproduced locally under a pty. Fix: derived names (raw emails) are stored as-is and never validated; validation applies only to user-typed names (`--profile`, `--save-as`), whose charset now also allows `+`. Regression tests added.

**greptile/veria on `user update domain` (valid):** the sync loop now repoints only profiles whose domain matched the roster entry's previous domain, so a same-email account on a different instance never has its token redirected to the new domain.

**greptile on the refresh path (valid):** the password/SRP login path now stores the refresh session scraped from the `jid` cookie (login2/MFA responses, and v3 login now scrapes it too), so the refresh wiring works for those sessions instead of only browser logins.

**Other failing checks, not related to this PR:**
- `govulncheck`: fails on GO-2026-6163 (panic in `github.com/pion/stun/v3`), a new advisory not in the workflow allowlist; go.mod is untouched here, so this fails for any PR scanned now. Needs a stun bump or allowlist entry in its own PR (the job also warns GO-2026-5942 can be dropped from the allowlist).
- `Agent End-to-End Testing`: transient proxy.golang.org download errors (`stream error: INTERNAL_ERROR`); CLI and PAM E2E in the same run passed. Should clear on re-run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

scott-ray-wilson:
@greptile review

scott-ray-wilson:
Verification script referenced in the description.

Runs in a throwaway `HOME`, so it never touches your real config or keychain. Needs an account with two organizations. Run it from a terminal: two of the checks assert that `profile pin` refuses to report success when its output is not captured, which requires a TTY on stdout and skips otherwise.

```sh
INFISICAL_BIN=./infisical \
VERIFY_DOMAIN=https://app.infisical.com \
VERIFY_EMAIL=you@example.com VERIFY_PASSWORD='...' \
VERIFY_ORG_A=<org-id> VERIFY_ORG_B=<org-id> \
./verify.sh
```

Last run against a local instance on the current head: **21 passed, 0 failed**.

<details>
<summary><code>verify.sh</code></summary>

```bash
#!/usr/bin/env bash
# Verifies the named-profiles branch against a live instance.
#
# Runs in a throwaway HOME, so your real ~/.infisical and keychain are never
# touched. Needs an account with TWO organizations.
#
#   INFISICAL_BIN=./infisical \
#   VERIFY_DOMAIN=http://localhost:8080 \
#   VERIFY_EMAIL=you@example.com VERIFY_PASSWORD='...' \
#   VERIFY_ORG_A=<org-id> VERIFY_ORG_B=<org-id> \
#   ./verify.sh
set -uo pipefail

BIN="${INFISICAL_BIN:?set INFISICAL_BIN to the built binary}"
DOMAIN="${VERIFY_DOMAIN:?set VERIFY_DOMAIN}"
EMAIL="${VERIFY_EMAIL:?set VERIFY_EMAIL}"
PASSWORD="${VERIFY_PASSWORD:?set VERIFY_PASSWORD}"
ORG_A="${VERIFY_ORG_A:?set VERIFY_ORG_A}"
ORG_B="${VERIFY_ORG_B:?set VERIFY_ORG_B}"

BIN="$(cd "$(dirname "$BIN")" && pwd)/$(basename "$BIN")"
SANDBOX="$(mktemp -d)"
trap 'rm -rf "$SANDBOX"' EXIT
export INFISICAL_DISABLE_UPDATE_CHECK=1 INFISICAL_DISABLE_MIGRATION_NOTICE=1

pass=0; fail=0
cli() { HOME="$SANDBOX" "$BIN" "$@"; }
check() { # check <description> <expected> <actual>
	if [ "$2" = "$3" ]; then printf '  ok    %s\n' "$1"; pass=$((pass+1))
	else printf '  FAIL  %s (expected %q, got %q)\n' "$1" "$2" "$3"; fail=$((fail+1)); fi
}
fresh() {
	rm -rf "$SANDBOX"/.infisical "$SANDBOX"/infisical-keyring
	cli vault set file >/dev/null 2>&1
	cli login --email "$EMAIL" --password "$PASSWORD" --organization-id "$ORG_A" \
		--domain "$DOMAIN" --plain --silent >/dev/null 2>&1
}
cfg() { python3 -c "import json,sys;print(json.load(open('$SANDBOX/.infisical/infisical-config.json')).get(sys.argv[1],''))" "$1"; }

echo "== login and migration =="
fresh
check "login creates a profile named after the email" "$EMAIL" "$(cli profile current --plain 2>/dev/null)"
check "legacy loggedInUserEmail still written" "$EMAIL" "$(cfg loggedInUserEmail)"
check "activeProfile set" "$EMAIL" "$(cfg activeProfile)"

rm -rf "$SANDBOX"/.infisical/infisical-config.json
mkdir -p "$SANDBOX"/.infisical
printf '{"loggedInUserEmail":"a@x.com","LoggedInUserDomain":"%s","loggedInUsers":[{"email":"a@x.com","domain":"%s"},{"email":"b@x.com","domain":"%s"}]}' "$DOMAIN" "$DOMAIN" "$DOMAIN" > "$SANDBOX"/.infisical/infisical-config.json
check "legacy roster migrates to one profile per account" "2" "$(cli profile list 2>/dev/null | sed 1d | grep -c .)"
check "no phantom profile from the mirror" "a@x.com" "$(cfg activeProfile)"

echo "== targeted login does not move the default =="
fresh
cli login --email "$EMAIL" --password "$PASSWORD" --organization-id "$ORG_B" \
	--domain "$DOMAIN" --profile second --plain --silent >/dev/null 2>&1
check "second profile exists" "2" "$(cli profile list 2>/dev/null | sed 1d | grep -c .)"
check "default profile unchanged by --profile login" "$EMAIL" "$(cli profile current --plain 2>/dev/null)"

echo "== org is a per-command dimension =="
before_org="$(cli profile current 2>/dev/null | awk -F': ' '/^Organization:/{print $2}')"
cli --org "$ORG_B" org list >/dev/null 2>&1
after_org="$(cli profile current 2>/dev/null | awk -F': ' '/^Organization:/{print $2}')"
check "--org leaves the profile default alone" "$before_org" "$after_org"
current_org() { cli org list 2>/dev/null | awk '$1=="*"{print $2}'; }
own_org="$(current_org)"
cli --org "$ORG_B" org list >/dev/null 2>&1
check "profile still reports its own org after --org (no token repointing)" "$own_org" "$(current_org)"

echo "== profile new =="
fresh
cli profile new third --org "$ORG_B" >/dev/null 2>&1
check "profile new creates without switching" "$EMAIL" "$(cli profile current --plain 2>/dev/null)"
cli profile new fourth --org "$ORG_B" --use >/dev/null 2>&1
check "--use switches the default" "fourth" "$(cli profile current --plain 2>/dev/null)"
check "--pin emits an export on stdout" "export INFISICAL_PROFILE=third" "$(cli profile new fifth --org "$ORG_B" >/dev/null 2>&1; cli profile pin third 2>/dev/null)"

echo "== selection precedence =="
fresh
cli login --email "$EMAIL" --password "$PASSWORD" --organization-id "$ORG_B" \
	--domain "$DOMAIN" --profile second --plain --silent >/dev/null 2>&1
mkdir -p "$SANDBOX"/proj/sub
( cd "$SANDBOX"/proj && HOME="$SANDBOX" "$BIN" profile bind second >/dev/null 2>&1 )
check "bound directory wins over the default" "second" "$(cd "$SANDBOX"/proj && HOME="$SANDBOX" "$BIN" profile current --plain 2>/dev/null)"
check "subdirectory inherits the binding" "second" "$(cd "$SANDBOX"/proj/sub && HOME="$SANDBOX" "$BIN" profile current --plain 2>/dev/null)"
check "env var beats the binding" "$EMAIL" "$(cd "$SANDBOX"/proj && HOME="$SANDBOX" INFISICAL_PROFILE="$EMAIL" "$BIN" profile current --plain 2>/dev/null)"
check "flag beats the env var" "second" "$(cd "$SANDBOX"/proj && HOME="$SANDBOX" INFISICAL_PROFILE="$EMAIL" "$BIN" --profile second profile current --plain 2>/dev/null)"
( cd "$SANDBOX"/proj && HOME="$SANDBOX" "$BIN" profile unbind >/dev/null 2>&1 )
check "unbind restores the default" "$EMAIL" "$(cd "$SANDBOX"/proj && HOME="$SANDBOX" "$BIN" profile current --plain 2>/dev/null)"

echo "== pin cannot report a false success =="
if [ -t 1 ]; then
	# stdout has to stay a terminal, otherwise it looks like eval captured it.
	cli profile pin second >/dev/tty 2>/dev/null
	check "bare pin exits non-zero when not eval'd" "1" "$?"
	check "pin through eval takes effect" "second" "$(eval "$(cli profile pin second 2>/dev/null)"; cli profile current --plain 2>/dev/null)"
else
	printf '  skip  bare pin check (stdout is not a terminal)\n'
fi

echo "== logout revokes, but not a shared session =="
fresh
cli login --email "$EMAIL" --password "$PASSWORD" --organization-id "$ORG_B" \
	--domain "$DOMAIN" --profile second --plain --silent >/dev/null 2>&1
token="$(cli user get token --plain --silent 2>/dev/null)"
cli logout --profile second >/dev/null 2>&1
code="$(curl -s -o /dev/null -w '%{http_code}' "$DOMAIN/api/v1/organization" -H "Authorization: Bearer $token")"
check "session survives while another profile uses it" "200" "$code"
cli logout --all >/dev/null 2>&1
code="$(curl -s -o /dev/null -w '%{http_code}' "$DOMAIN/api/v1/organization" -H "Authorization: Bearer $token")"
check "logout --all revokes the session server-side" "revoked" "$([ "$code" = "200" ] && echo "still valid" || echo "revoked")"

printf '\n%d passed, %d failed\n' "$pass" "$fail"
[ "$fail" -eq 0 ]
```

</details>

scott-ray-wilson:
@veria-ai review

scott-ray-wilson:
@greptile review

scott-ray-wilson:
@greptile review

scott-ray-wilson:
@greptile review

packages/api/api.go
victorvhs017, line 1: Unrelated to the file, but some considerations I had while testing this:

- Why the organization name and not the organization slug in the generated profile name?
- Why not use the org in the first generated profile name? After you have logged in to a few orgs it can get hard to remember the org without listing
- With these auto-generated names, a rename command could be useful here, instead of deleting and setting up a new profile.
- Should login status https://infisical.com/docs/cli/commands/login#infisical-login-status:check-the-active-user-session show profile information?

packages/cmd/init.go
```diff
@@ -57,64 +58,66 @@ var initCmd = &cobra.Command{
...
+
+			// An --org override is per command, so a project linked under it would
+			// not resolve on later runs that use the profile's default.
+			if userCreds.OrganizationSource != util.OrgSourceProfileDefault && userCreds.Profile.OrganizationID != "" && userCreds.OrganizationID != userCreds.Profile.OrganizationID {
+				profileOrg := userCreds.Profile.OrganizationName
+				if profileOrg == "" {
+					profileOrg = userCreds.Profile.OrganizationID
+				}
+				util.PrintWarning(fmt.Sprintf("Profile '%s' defaults to organization %s, so later commands here will not find this project unless you pass --org again. Run [infisical profile set-org %s] to make it the default.", userCreds.ProfileName, profileOrg, orgDisplay))
+			}
```
victorvhs017, line 120: I'm a bit concerned that this warning is not enough here.

They can be silenced by --silent, and the errors from the command won't hint that the issue is this mismatch between the profile org and the project from another org.

I think we should raise an error in this case.

packages/cmd/org.go
```diff
@@ -0,0 +1,279 @@
...
+			verifyMFAresponse, mfaErrorResponse, requestError := api.CallVerifyMfaToken(httpClient, api.VerifyMfaTokenRequest{
+				Email:     email,
+				MFAToken:  mfaVerifyCode,
+				MFAMethod: tokenResponse.MfaMethod,
+			})
+			if requestError != nil {
+				return "", requestError
+			} else if mfaErrorResponse != nil {
+				if mfaErrorResponse.Context.Code == "mfa_invalid" {
+					msg := fmt.Sprintf("Incorrect, verification code. You have %v attempts left", 5-i)
```
victorvhs017, line 248: We use this same "5" in the login.go, we should put it in a shared const to keep them synced.

packages/cmd/profile.go
```diff
@@ -0,0 +1,642 @@
...
+// requireShellCapture stops a shell-mutating command that was run bare, and
+// shows the form that actually works, rather than reporting a success that did
+// not happen.
+func requireShellCapture(invocation string) {
+	if shellOutputIsCaptured() {
+		return
+	}
+	util.PrintlnStderr(fmt.Sprintf("This command works by printing a shell statement, so it only takes effect when the shell reads it:\n\n    eval \"$(%s)\"\n\nNothing has been changed. Tip: add a shell alias if you use this often.", invocation))
+	os.Exit(1)
+}
```
victorvhs017, line 66: This may not work depending on the OS/shell/SSH setup. For example, PowerShell wouldn’t support the eval "$(…)" flow, and non-TTY SSH/scripts can look successful without actually pinning.

We should update the error message (and profile pin help) to address these scenarios too, as the pin command would be unusable for them.

We should also mention the PowerShell/manual env-var alternative.

packages/cmd/profile.go
```diff
@@ -0,0 +1,642 @@
...
+		util.PrintlnStdout("Selected via:", selectedVia)
+		if resolved.ShadowedName != "" {
+			util.PrintlnStdout("Overrides binding:", fmt.Sprintf("%s (bound at %s). Run [eval \"$(infisical profile unpin)\"] to use it.", resolved.ShadowedName, resolved.ShadowedScopeDir))
+		}
+
+		if !found {
+			util.PrintlnStdout("Status: profile does not exist. Run [infisical login --profile " + resolved.Name + "] to create it.")
+			return
+		}
+
```
victorvhs017, line 172: I'm getting a bit confused by profile login and profile new. I understand the difference and usage of each, but when it comes to what feels more natural, I would always try to create a new profile using the new command.

login for me -> logging into a profile, not creating one
new for me -> actually creating a new profile from the ground up

Maybe some renaming here; what about "configure" instead of login? or "setup"?
And the new feels more like "copy" or "duplicate" to me. "fork"/"clone" maybe?

packages/util/profile.go
```diff
@@ -0,0 +1,777 @@
...
+	ShadowedName     string
+	ShadowedScopeDir string
+}
+
+func ValidateProfileName(name string) error {
+	if !profileNamePattern.MatchString(name) {
+		return fmt.Errorf("invalid profile name '%s': use letters, digits, and the characters @ . _ + - (must start with a letter or digit)", name)
+	}
+	return nil
+}
```
victorvhs017, line 58: We should validate length because names too long can fail to store in the keyring

packages/util/credentials.go
```diff
@@ -58,61 +103,191 @@ func GetUserCredsFromKeyRing(userEmail string) (credentials models.UserCredentia
...
+func applyOrgOverride(details *LoggedInUserDetails, selector string, selectorSource string) error {
+	profile := details.Profile
+
+	// Already on the requested organization: nothing to do, and no API calls.
+	// Only an id match is trusted here, since a name or slug could belong to a
+	// different organization and would skip the exchange wrongly.
+	if OrgMatchTier(selector, profile.OrganizationID, "", "") == orgMatchID {
+		details.OrganizationSource = selectorSource
+		return nil
+	}
```
victorvhs017, line 202: I was checking how much this would increase the API calls volume, and in the default path it's fine, but scripts using --org <name> against the profile’s default org may list orgs every run (only id match skips today), because org list APIs aren’t cached

packages/cmd/init.go
```diff
@@ -57,64 +58,66 @@ var initCmd = &cobra.Command{
...
-		}
+			// The session token is now scoped to the selected organization; record
+			// it on the profile this invocation resolved to so later commands in
+			// this project don't have to ask again.
+			userCreds.UserCredentials.JTWToken = newSessionToken
+			orgID, subOrgID := util.ParseTokenOrgClaims(newSessionToken)
+			if orgID == "" {
+				orgID = pickedOrgID
+			}
+			selectedOrgID = orgID
```
victorvhs017, line 88: For sub-organization sessions, the JWT’s organizationId is the root org, not the sub-org the user selected. init uses that root id to filter projects, but sub-org–scoped tokens return projects tagged with the sub-org id, so the filter finds nothing and init aborts even when projects exist. The profile stores subOrganizationId, but project filtering doesn’t use it.

packages/cmd/init.go
```diff
@@ -57,64 +58,66 @@ var initCmd = &cobra.Command{
...
 
-		selectedOrgID, selectedSubOrgName, err := pickOrganization(httpClient, "Which Infisical organization would you like to select a project from?", userCreds.UserCredentials.Email)
-		if err != nil {
-			util.HandleError(err, "Unable to select organization")
-		}
+		// The profile already carries an organization (and --org can retarget it
+		// for this command), so don't ask again. Only fall back to the picker
+		// when the profile has no organization recorded, which happens for
+		// sessions migrated from a CLI that predates profiles.
+		selectedOrgID := userCreds.OrganizationID
```
victorvhs017, line 65: same issue here, the selectedOrgID wouldn't be the suborgID

packages/util/credentials.go
victorvhs017, line 202: Same issue here

packages/cmd/org.go
```diff
@@ -0,0 +1,279 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		details := requireUserSession()
+
+		httpClient, err := util.GetRestyClientWithCustomHeaders()
+		if err != nil {
+			util.HandleError(err, "Unable to get resty client with custom headers")
+		}
+		httpClient.SetAuthToken(details.UserCredentials.JTWToken)
+
+		currentOrgID := details.OrganizationID
```
victorvhs017, line 49: same here

packages/cmd/user.go
victorvhs017, line 1: getLoggedInUsers and LoggedInUsersPrompt are now dead code, right?

packages/util/credentials.go
```diff
@@ -58,61 +103,191 @@ func GetUserCredsFromKeyRing(userEmail string) (credentials models.UserCredentia
...
+	// and the exchange. Pick the strongest match rather than the first, so a
+	// cached entry matching only by name cannot shadow one matching by id.
+	bestCached := models.CachedOrgSession{}
+	bestTier := 0
+	for _, cached := range details.UserCredentials.OrgTokens {
+		tier := OrgMatchTier(selector, cached.OrgID, cached.OrgSlug, cached.OrgName)
+		if tier > bestTier && !IsJWTExpired(cached.Token) {
+			bestCached, bestTier = cached, tier
 		}
+	}
```
victorvhs017, line 214: We resolve by bare name (Research) but store OrgName as "Acme / Research". Repeat `--org Research` never hits orgTokens, so every command re-lists orgs and re-exchanges.

packages/util/credentials.go
victorvhs017, line 1: I log in once (default org Acme / Production). Over time, I run:

```
infisical --org staging secrets ...
infisical --org qa secrets ...
infisical --org client-a secrets ...
infisical --org client-b secrets ...
# ... or INFISICAL_ORG=client-c in CI
```

Each new non-default org adds a full session JWT to orgTokens in the same keyring blob. After ~5 such orgs (on macOS, with our token sizes), the keyring write can fail; commands still work for that run, but the cache never sticks and every --org re-lists/re-exchanges.

victorvhs017 (changes requested):

packages/api/api.go
scott-ray-wilson: All four are in now.

**Slug, not name.** Profiles are named `<email>--<org-slug>`, e.g. `scott@example.com--acme-x4k2`. The slug is stable and already URL-safe. If an instance reports no slug we fall back to the slugified name, then to a short piece of the id.

**Org in the first name too.** Every login gets the suffix now, including the first one, so you can tell the tenant from the name without listing.

**Rename.** Added `infisical profile rename <old> <new>`. It moves the stored session and the cached org sessions, and updates the default profile and any directory bindings. A pinned terminal keeps pointing at the old name, so it tells you to pin again.

**`login status`.** It now shows the profile and how it was selected, the org name next to its id, and `Organization via` when `--org` is in play. Same fields in `--json` (`profile`, `profileSource`, `organizationName`).

One trade-off worth flagging: a profile that is not named after its email does not publish the legacy `loggedInUserEmail` field. An older CLI binary reading the same config will ask for a fresh login instead of loading the wrong profile's token. Scott and I decided that is the right call, but it does mean a downgrade after a fresh login is no longer seamless.

packages/cmd/init.go
scott-ray-wilson: Agreed. It is an error now, not a warning.

`init` exits 1 without writing anything, and names the two durable fixes: make it the profile's default with `profile set-org`, or keep both orgs by creating a second profile and `profile bind`ing it to the directory.

packages/cmd/org.go
scott-ray-wilson: Done, both loops use a shared `mfaMaxAttempts`.

Unrelated thing I hit while testing this with MFA turned on: a wrong code never prints "You have N attempts left". The CLI matches on `context.code == "mfa_invalid"`, but the backend now returns a plain 500 with no code, so the CLI just re-prompts with no explanation. That is pre-existing on `main`, not from this PR. Happy to file it separately.

packages/cmd/profile.go
scott-ray-wilson: Done. The error and `profile pin --help` now spell out the alternatives:

```
$env:INFISICAL_PROFILE = 'globex'   # PowerShell
set INFISICAL_PROFILE=globex        # cmd.exe
export INFISICAL_PROFILE=globex     # scripts and CI, or just use --profile
```

I also fixed the false-success problem you implied. Capturing stdout does not prove a shell evaluated it, it could be a pipe or a file, so the message now says the profile is pinned *once the shell evaluates the printed export* rather than claiming it already happened.

packages/cmd/profile.go
scott-ray-wilson: You are right that `login --profile x` reads like "log into profile x", not "create it".

I kept the command names and split the two meanings onto separate flags:

- `login --save-as <name>` creates the profile, or replaces its session. This is the "set it up" verb.
- `login --profile <name>` only signs back in to a profile that already exists. It keeps the account, instance and org, so you do not pass `--organization-id` again and you get no hosting prompt. If the profile does not exist it fails and points you at `--save-as`.
- `profile new` is unchanged: reuse the login you already have for another org.

So `--profile` now means the same thing on `login` as everywhere else, which is "use this profile", and creation has its own flag.

I did not rename the commands to `configure`/`setup` or `clone`/`fork`. The confusing part was the flag rather than the verb, and renaming would break people already scripting against them. Say the word if you would still rather have the rename.

packages/util/profile.go
scott-ray-wilson: Done. Names you type are capped at 64 characters, checked in `login --save-as`, `profile new` and `profile rename`. Names derived from your email are left alone, since those have always been valid keyring keys.

packages/util/credentials.go
scott-ray-wilson: Fixed, and this turned out to be the more interesting half of the comment.

`--org` is now resolved against what the profile already knows before any API call. Profiles record the org slug at login, so your own org matches locally by id, slug or name. Orgs you have used before are kept in a small index on the profile and match the same way.

Measured against a local instance by counting requests in the access log:

| `--org` selector | listing calls | exchanges |
|---|---|---|
| your own org, by id / slug / name | 0 (first use included) | 0 |
| a new org, first use | 1 | 1 |
| that same org afterwards | 0 | 0 |

So the script pattern you described costs nothing after the first run.

One trade-off: a name that matches your own org now short-circuits locally, so we no longer catch the case where you also belong to a different org with the same name and have never selected it. Ids and slugs are unique per instance, so those are unaffected.

packages/cmd/init.go
scott-ray-wilson: Good catch, this was real and it was the worst bug in the PR.

I confirmed the premise against a local instance first: a project created with a sub-org token comes back tagged with the sub-org id, so filtering by the token's root id found nothing and `init` aborted.

`Profile` now has `ScopedOrganizationID()`, which is the sub-org when the session is scoped to one and the root otherwise. `init` filters projects with that.

Verified end to end against a seeded sub-org with its own project: `init` on a sub-org profile lists and links that project, and a root-org profile still sees only its own projects and not the sub-org's.

packages/cmd/init.go
scott-ray-wilson: Same fix. `selectedOrgID` is the scoped id now, including on the picker path that still runs for older profiles with no org recorded yet.

packages/util/credentials.go
scott-ray-wilson: Same fix. The `--org` comparison uses the scoped id, so `--org <root-id>` on a sub-org profile now does the exchange instead of matching and silently doing nothing.

packages/cmd/org.go
scott-ray-wilson: Same fix. `org list` reads `subOrganizationId` from the token first, so the `*` lands on the sub-org row rather than on its parent.

packages/cmd/user.go
scott-ray-wilson: Yes. Both removed.

packages/util/credentials.go
scott-ray-wilson: Fixed. A selector now matches a sub-org's own name as well as the full `Acme / Research` label, so `--org Research` hits the cache instead of re-listing every time.

Verified by counting requests: the first `--org Research` does one lookup and one exchange, and every run after that does neither.

packages/util/credentials.go
scott-ray-wilson: Fixed by not storing the tokens there at all.

Each cached org session is now its own keyring entry, `org-session:<profile>:<orgId>`. The profile keeps only the id, name and slug, which is all a selector needs to match on. So the profile's own entry does not grow with the number of orgs you use.

Checked with three cached orgs: three separate keyring entries, and the profile's own entry stayed around 1 KB, well under the macOS limit. `logout`, `profile delete`, `profile rename`, `user update domain` and `reset` all clean up the extra entries.

packages/util/profile.go
```diff
@@ -0,0 +1,476 @@
...
+	}
+
+	configFile.ActiveProfile = name
+	syncLegacyLoginFields(configFile, profile)
+	return nil
+}
+
+func syncLegacyLoginFields(configFile *models.ConfigFile, profile models.Profile) {
+	configFile.LoggedInUserEmail = profile.Email
+	configFile.LoggedInUserDomain = profile.Domain
```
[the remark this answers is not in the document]
scott-ray-wilson: This is handled. The legacy `loggedInUserEmail` / `LoggedInUserDomain` fields are only written when the profile's name equals its email, which is exactly the case where the keyring entry an older CLI would load is that same profile's entry.

Any other profile name leaves the legacy pointer empty, so an older binary asks for a fresh login rather than loading one profile's token and sending it to another profile's instance.

Since profiles created by this branch are now named `<email>--<org-slug>`, that is the common case.

packages/util/credentials.go
```diff
@@ -58,61 +98,187 @@ func GetUserCredsFromKeyRing(userEmail string) (credentials models.UserCredentia
...
+				// since it may be a proxy or tunnel for the same instance. Surface
+				// the mismatch once so a wrong-instance 401 is explainable.
+				if profileURL != config.INFISICAL_URL {
+					domainMismatchNoticeOnce.Do(func() {
+						PrintWarning(fmt.Sprintf("The explicitly set domain '%s' differs from profile '%s' domain '%s'. Requests will go to the explicitly set domain.", DisplayDomain(config.INFISICAL_URL), profile.Name, DisplayDomain(profileURL)))
+					})
+				}
 			} else {
-				return LoggedInUserDetails{}, fmt.Errorf("failed to fetch credentials from keyring because [err=%s]", err)
+				config.INFISICAL_URL = profileURL
```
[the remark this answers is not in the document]
scott-ray-wilson: This is handled. When `--domain` or `INFISICAL_DOMAIN` is set explicitly and does not match the resolved profile's own instance, credential resolution fails with `ErrProfileDomainMismatch` and the token is never attached to a request. The error names both instances and tells you to log in to the other one as a separate profile.

`login` treats that error as "no usable session yet" rather than a hard failure, so you can still log in to a new instance. Signing back in with `login --profile <name> --domain <other>` is refused for the same reason.

packages/cmd/login.go
```diff
@@ -220,48 +279,80 @@ var loginCmd = &cobra.Command{
...
-			orgID, subOrgID := util.ParseTokenOrgClaims(userCredentialsToBeStored.JTWToken)
-			orgName := util.OrgDisplayName(userCredentialsToBeStored.JTWToken, orgID, subOrgID)
-
-			existingConfig, err := util.GetMigratedConfigFile()
-			if err != nil {
-				util.HandleError(err, "Unable to read the Infisical config file")
+			// The browser flow scopes the session to whatever organization the
+			// browser had selected. Signing back in to a profile keeps its
+			// organization, so re-scope the session when the two differ.
+			if wantOrg := target.profile.ScopedOrganizationID(); target.reauth && wantOrg != "" {
```
victorvhs017, line 324: Re-scoping to the profile org runs before the wrong-account check. We can have something like this:

```
# Setup: profile belongs to user A
infisical login --save-as user-a-profile   # as victor@...

# Try re-auth as user B into user A's profile
infisical login --profile user-a-profile   # sign in as other@... in browser

# Expect (designed):  "Profile belongs to victor@..., but you signed in as other@..."
# Actual (2.10b):     403 / "User does not have access to the organization"
```

packages/util/profile.go
victorvhs017, line 630: **Self-hosted gateway/relay/kmip lose instance fallback (regression)**

Clearing `LoggedInUserDomain` when `profile.Name != profile.Email` is intentional for old-CLI compat, but with `DeriveProfileName` that's now the normal case (e.g. victor@...--admin-org-we-68). The instance URL still lives on `profile.Domain`.

Affected call sites still fall back to LoggedInUserDomain only:

```
packages/cmd/gateway.go:264-265
packages/cmd/relay.go:67-68
packages/cmd/kmip.go:83-84 (resolveKmipDomain)
Self-hosted users without --domain and without a stored gateway/relay/kmip domain silently get app.infisical.com instead of their logged-in instance.

```
Suggestion: Replace the `LoggedInUserDomain` fallback at those three sites with the active profile's Domain (via `ResolveProfile` / `FindProfile`), same as the rest of the profile-aware CLI.

victorvhs017, line 801: **Sub-orgs under one root collapse into a single profile**

The reuse check compares `profile.OrganizationID == orgID` (root only). On login, `login.go:329` passes the root orgID while SubOrganizationID is stored separately on the profile.

Example

```
Login 1 — pick sub-org "test 1"
  → profile: victor@...--admin-org-we-68
  → OrganizationID:    root-id
  → SubOrganizationID: test-1-id
  → keyring token scoped to test 1

Login 2 — pick sub-org "test 2"
  → DeriveProfileName: same email, same domain, same root orgID
  → reuse check matches Login 1's profile ✅ (wrong!)
  → same name: victor@...--admin-org-we-68
  → PersistLoginProfile OVERWRITES profile + keyring
  → test 1 session is gone
```

victorvhs017, line 601: `UpsertProfile` replaces the whole profile; `login.go:349` omits `OrgSessions`. Re-login wipes the index but leaves `org-session:*` keyring entries orphaned, `logout/delete/reset` can't find them. Preserve existing `OrgSessions` on upsert.

```
Before re-login:
  config:   orgSessions: [{orgId: B, ...}, {orgId: C, ...}]
  keyring:  globex-work
            org-session:globex-work:org-B
            org-session:globex-work:org-C

After re-login:
  config:   orgSessions: []          ← wiped
  keyring:  globex-work              ← updated
            org-session:...:org-B   ← orphaned
            org-session:...:org-C   ← orphaned

```

packages/util/logout.go
```diff
@@ -121,8 +144,13 @@ func LogoutProfiles(configFile models.ConfigFile, targetNames []string, localOnl
...
 
-		creds, err := GetUserCredsFromKeyRing(name)
-		if err != nil {
+		profile, found := FindProfile(configFile, name)
+		if !found {
+			profile = models.Profile{Name: name}
+		}
+
+		tokens, ok := profileTokens(profile)
+		if !ok {
```
victorvhs017, line 178: `profileTokens` returns `ok=false` when the main keyring entry is missing, so `LogoutProfiles` skips `ClearStoredSession`. Cached `org-session:*` tokens are never deleted/revoked. It should still clean up org sessions from `orgSessions` when the index exists.

packages/util/profile.go
victorvhs017, line 753: `profile delete` handles `SharedWith/RevokeErr/Revoked` but not `LocalErr`. Keyring delete failures are silent while the config entry is removed, user sees "Deleted profile" but tokens may remain. `logout` warns on `LocalErr` at `logout.go:83`; same handling needed here.

victorvhs017 (changes requested):

packages/util/profile.go
```diff
@@ -0,0 +1,1047 @@
...
+	// superseded session and fail at request time. Drop the keyring entries and
+	// the index together, so neither outlives the other.
+	if existing, found := FindProfile(configFile, profile.Name); found {
+		for _, ref := range existing.OrgSessions {
+			if err := DeleteOrgSessionToken(existing.Name, ref.OrgID); err != nil {
+				log.Debug().Err(err).Str("profile", existing.Name).Str("organization", ref.OrgID).Msg("unable to remove a cached organization session")
+			}
+		}
+	}
+	profile.OrgSessions = []models.OrgSessionRef{}
```
victorvhs017: `PersistLoginProfile` always wipes all `OrgSessions` keyring entries. Correct for login (new session voids old cached org tokens), but `profile set-org` also uses this, it only exchanges the existing session, so other cached `--org` sessions are still valid. Makes `org.go:186-192` dead code and deletes tokens locally without server revocation.

Example:

```
Profile globex-work:
  default org: Admin Org
  cached --org sessions: Main Admin Org, test-1 sub-org

You run: infisical profile set-org test-1

Expected:
  default → test-1
  remove cached test-1 (now redundant)
  KEEP cached Main Admin Org

Actual:
  default → test-1
  ALL cached org sessions deleted from keyring + index wiped
```

victorvhs017: On failed `DeleteOrgSessionToken`, errors are debug-only, but `orgSessions` is wiped anyway, leaving the keyring entry orphaned with no index to find/revoke it. Comment says drop both together; on error it drops only the index.

Example:

```
1. Profile has cached org sessions for B and C
2. User re-logs in
3. Delete org-B keyring entry → fails (keychain locked)
4. Delete org-C → succeeds
5. Index wiped to []
6. org-B token orphaned in keyring, invisible to CLI
```

victorvhs017 (changes requested):

packages/cmd/login.go
scott-ray-wilson: Fixed in af89c55.

The account is now checked as soon as it is known, which is before either login path applies the profile's organization. Your repro was the browser flow; while testing I found the password flow reached `select-organization` even earlier, through `cliDefaultLogin`'s default `--organization-id`, so both paths now check via a shared `exitOnLoginAccountMismatch`.

Verified against two real accounts on a local instance (a second seeded user that is not a member of the first's org). Before the fix that reproduces your exact symptom, `403 / User does not have access to the organization`; after it, the intended message.

packages/util/profile.go
scott-ray-wilson: Fixed in f4e46e7.

Confirmed as you described: `DeriveProfileName` appends an org suffix, so `Name != Email` is the normal case and `syncLegacyLoginFields` clears the mirror on essentially every login. Added `util.ActiveProfileDomain()` (resolve the profile, fall back to the legacy field for pre-profile configs) and used it at all three sites.

Worth flagging for your own retest: the failure is invisible from the error text. Pre-fix `relay start` printed an identical "Invalid enrollment token" while talking to app.infisical.com. `kmip start` shows it plainly: before, `POST https://app.infisical.com/...`; after, `POST http://localhost:8080/...`.

scott-ray-wilson: Fixed in 8416960.

Exactly as you traced it. The naming was already correct (`DescribeSessionOrg` resolves the scoped org, so the suffixes differ); only the reuse check compared the root. It now compares `profile.ScopedOrganizationID()` against the scoped id, and login passes the sub-org when there is one.

Live check: logging into the Research sub-org and then the Acme root left one profile before the fix (the sub-org profile silently overwritten) and two after.

scott-ray-wilson: Fixed in 66cf327, then corrected in edeaeec and e2f57aa after your follow-up.

`UpsertProfile` keeps `OrgSessions` when a caller does not carry them, so a login rebuilding the profile from the session alone no longer strands the keyring entries.

packages/util/logout.go
scott-ray-wilson: Fixed in 840b949.

`profileTokens` now accumulates instead of bailing on the first miss, so a missing main entry no longer skips `ClearStoredSession`. It also fixes a second-order case the comment did not mention: the same early return dropped those sessions from the `retained` set, so a session another profile still relied on could be revoked out from under it.

Live check: with the main entry removed, logout now revokes and deletes the cached org session; before, it reported "had no stored session" and left the token live.

packages/util/profile.go
scott-ray-wilson: Fixed in a6f1532.

Handled the same way `logout` does. The message also says the leftover entries must be removed by hand, because `RemoveProfile` drops the index that named them, so nothing can find them afterwards.

One judgement call for you: the profile is still deleted when the keyring fails, on the grounds that users expect delete to delete. The alternative is aborting so a retry can still find the entries; happy to switch if you prefer that.

packages/util/profile.go
scott-ray-wilson: Fixed in edeaeec. You were right, and I measured it: all tokens from one login share a session id, and the cached token still returns 200 after `set-org`, so they were never void.

Rather than special-casing the caller, the purge is now gated on the session actually changing (`orgSessionsSurvive`). That covers `set-org`, `init`, and a re-login the server answers with the existing session, all of which were destroying valid cache. `org.go:186-192` is live code again and is the only cache change `set-org` makes.

Also worth noting: "deletes tokens locally without server revocation" is accurate, but revocation would be the wrong remedy here, since every cached org token carries the main session's id and revoking it would sign the whole profile out. The harm was lost cache, not a dangling credential.

scott-ray-wilson: Fixed in e2f57aa.

`purgeOrgSessions` now returns the entries whose keyring token could not be deleted, so they stay indexed and findable, and the failure is logged at warn rather than debug. The comment matches the code now.

greptile then pointed out the other side of this: a kept entry is also a servable one. That is fixed in 45229a2, which requires a cached token to belong to the session in hand before it is sent.

packages/util/profile.go
```diff
@@ -0,0 +1,1081 @@
...
+// entries later, so forgetting one would strand the token with nothing left to
+// reach it by.
+func purgeOrgSessions(profile models.Profile) []models.OrgSessionRef {
+	kept := []models.OrgSessionRef{}
+	for _, ref := range profile.OrgSessions {
+		if err := DeleteOrgSessionToken(profile.Name, ref.OrgID); err != nil {
+			log.Warn().Err(err).Str("profile", profile.Name).Str("organization", ref.OrgID).
+				Msg("unable to remove a cached organization session; keeping it on the profile so it can still be removed")
+			kept = append(kept, ref)
+			continue
```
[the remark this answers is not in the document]
scott-ray-wilson: Fixed in 45229a2. Confirmed, and the cause is older than the retention change: any cached token can outlive its session, since `applyOrgOverride` gated only on expiry.

Gated on session membership instead of patching the purge, so the entry stays findable for cleanup (which victorvhs017 asked for) and can never be served. An unusable cached token now falls through to the fresh exchange the code already had.

The live check nearly passed vacuously, which is worth knowing if you retest: the pre-fix binary does return a JWT, just a dead one. Only asserting the token works against the server caught it (404 before, 200 after).

packages/util/profile.go
```diff
@@ -0,0 +1,1081 @@
...
+	if previous, err := GetUserCredsFromKeyRing(profile.Name); err == nil {
+		previousSessionID = ParseTokenSessionID(previous.JTWToken)
+	}
+
+	// Deliberately no name validation here: derived names are raw account
+	// emails (which may contain any RFC-legal character) and have always been
+	// valid keyring keys. Rejecting them would block login entirely. Name
+	// validation applies only where users type a name (--profile, profile new,
+	// profile rename), at the command layer.
+	if err := StoreUserCredsInKeyRing(profile.Name, userCred); err != nil {
```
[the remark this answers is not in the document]
scott-ray-wilson: Fixed in d5dc793.

The previous credentials are restored if anything after the keyring write fails, and the entry is removed outright when the profile had none, so a new token is never left paired with stale instance metadata. Reordering was not the answer, since writing the config first just inverts the leak.

packages/util/logout.go
```diff
@@ -0,0 +1,266 @@
...
+// session. The process-wide domain is restored afterwards.
+func LogoutProfilesAcrossDomains(configFile models.ConfigFile, targetNames []string, localOnly bool) []LogoutResult {
+	originalURL := config.INFISICAL_URL
+	defer func() { config.INFISICAL_URL = originalURL }()
+
+	byDomain := map[string][]string{}
+	for _, name := range targetNames {
+		domain := originalURL
+		if profile, found := FindProfile(configFile, name); found && profile.Domain != "" {
+			domain = AppendAPIEndpoint(profile.Domain)
```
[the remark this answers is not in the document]
scott-ray-wilson: Leaving this one open for a maintainer to weigh in, because I do not think it is actionable as described.

Logout sends the session token to the profile's own stored domain, which is where every other authenticated command already sends it: `credentials.go` assigns `config.INFISICAL_URL = profileURL` during credential resolution. The logout path introduces no new trust boundary.

The adjacent guard shows this risk was already considered. When an explicit `--domain` differs from the profile's, the CLI refuses to send the token at all, in as many words: "this profile's session was issued by a different instance and must not be sent there: it would hand a valid bearer token to whoever runs that host."

The attack also needs local write access to `~/.infisical/infisical-config.json`, and an attacker with that can already redirect every command's token through the same field. That makes it a local config-trust property rather than SSRF, since no untrusted remote input drives the request.

Hardening config-domain trust seems worthwhile, but as its own change covering every call site rather than bolted onto logout.

packages/cmd/gateway.go
```diff
@@ -148,8 +148,8 @@ var gatewayStartCmd = &cobra.Command{
 			config.INFISICAL_URL = util.AppendAPIEndpoint(flagDomain)
 		} else if storedDomain, _ := gatewayv2.LoadStoredDomain(gatewayName); storedDomain != "" {
 			config.INFISICAL_URL = util.AppendAPIEndpoint(storedDomain)
-		} else if configFile, cfgErr := util.GetConfigFile(); cfgErr == nil && configFile.LoggedInUserDomain != "" {
-			config.INFISICAL_URL = util.AppendAPIEndpoint(configFile.LoggedInUserDomain)
+		} else if profileDomain := util.ActiveProfileDomain(); profileDomain != "" {
+			config.INFISICAL_URL = util.AppendAPIEndpoint(profileDomain)
```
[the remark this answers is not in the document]
scott-ray-wilson: Same assessment as the logout thread, and the diff here makes it clearer.

The line this replaces was `configFile.LoggedInUserDomain`, read from the same `~/.infisical/infisical-config.json`. Both before and after, gateway reads a locally editable domain out of that one file; the change only swaps which field of it is used, so it adds no trust that was not already there. The `LoggedInUserDomain` fallback is in fact still the last resort inside `ActiveProfileDomain()`.

The change was needed because that legacy field is only written for a profile named after its own account email, which derived names are not, so self-hosted users were silently falling back to app.infisical.com (victorvhs017 reported this above).

If we want an allowlist or trusted-origin check on stored domains I am happy to write it, but it should cover every site that reads a domain from local config, not just these three.

packages/util/logout.go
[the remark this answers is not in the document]
scott-ray-wilson: Maintainer reviewed: declining, resolving as won't-fix.

Sending a profile's token to that profile's own stored instance is the CLI's existing and intended behavior, not something this path introduces, and the threat presupposes local write access to the config that would compromise every command equally.

If we do want to constrain which stored domains the CLI will talk to, that is a separate change covering every reader of a config-supplied domain. Happy to pick it up as its own PR.

packages/cmd/gateway.go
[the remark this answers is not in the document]
scott-ray-wilson: Maintainer reviewed: declining, resolving as won't-fix.

Same reasoning as the logout thread. This line swapped one locally editable field of `~/.infisical/infisical-config.json` for another in the same file, so the trust level is unchanged, and the swap was necessary to stop self-hosted users silently falling back to app.infisical.com.

A trusted-origin check on stored domains would be a separate change covering every such reader rather than these three sites.

scott-ray-wilson:
@greptile review

victorvhs017 (approved):

PR #374, opened 2026-08-26
feat: add lifecycle.use-latest so certificate-id follows server-side renewals

carlosmonastyrski:
# Description 📣

Extends the certificate agent's existing `certificate-id` distribution mode with `lifecycle.use-latest`, so a fleet of hosts sharing one centrally managed certificate automatically picks up renewals made in Infisical without a config change or a restart. The agent resolves the current certificate from the server in one call regardless of how far the chain has moved on, rewrites files and runs `on-renewal` only when the certificate it resolves to actually changes (including across a restart, where it compares against what is on disk rather than in memory), reports revoked and expired certificates through `on-failure` while continuing to serve the last good certificate, and refuses to overwrite a certificate when it cannot also replace the matching private key. `renew-before-expiry` warns and is ignored in this mode.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [x] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/certificate-agent-config.yaml
+++ b/certificate-agent-config.yaml
@@ -56,12 +56,16 @@ certificates:
         permission: "0644"
         omit-root: true
 
-  # Fetch an existing certificate by ID. Skips issuance entirely.
-  # If private-key.path is configured but the certificate has no private key
-  # (e.g. ACME-issued), a warning is logged and the file is skipped.
+  # Distribution mode: deliver a certificate that already exists in Infisical instead of issuing one,
+  # so several hosts can share it. https://infisical.com/docs/integrations/platforms/certificate-agent
   # - certificate-id: "00000000-0000-0000-0000-000000000000"
   #   lifecycle:
   #     status-check-interval: "6h"
+  #     use-latest: true # Also deliver renewals made in Infisical
+  #   post-hooks:
+  #     on-renewal:
+  #       command: "systemctl reload nginx"
+  #       timeout: 30
   #   file-output:
   #     certificate:
   #       path: "./certs/existing/certificate.crt"
--- a/certificate-agent-config.yaml
+++ b/certificate-agent-config.yaml
@@ -61,7 +61,7 @@ certificates:
   # - certificate-id: "00000000-0000-0000-0000-000000000000"
   #   lifecycle:
   #     status-check-interval: "6h"
-  #     use-latest: true # Also deliver renewals made in Infisical
+  #     replace-on-renewals: true # Also deliver renewals made in Infisical
   #   post-hooks:
   #     on-renewal:
   #       command: "systemctl reload nginx"
--- a/e2e/agent/agent_helpers.go
+++ b/e2e/agent/agent_helpers.go
@@ -591,7 +591,7 @@ type agentCertificateAttributes struct {
 type agentCertificateLifecycle struct {
 	RenewBeforeExpiry   string `yaml:"renew-before-expiry,omitempty"`
 	StatusCheckInterval string `yaml:"status-check-interval,omitempty"`
-	UseLatest           bool   `yaml:"use-latest,omitempty"`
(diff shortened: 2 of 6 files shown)


certificate-agent-config.yaml
```diff
@@ -56,12 +56,16 @@ certificates:
...
 
-  # Fetch an existing certificate by ID. Skips issuance entirely.
-  # If private-key.path is configured but the certificate has no private key
-  # (e.g. ACME-issued), a warning is logged and the file is skipped.
+  # Distribution mode: deliver a certificate that already exists in Infisical instead of issuing one,
+  # so several hosts can share it. https://infisical.com/docs/integrations/platforms/certificate-agent
   # - certificate-id: "00000000-0000-0000-0000-000000000000"
   #   lifecycle:
   #     status-check-interval: "6h"
+  #     use-latest: true # Also deliver renewals made in Infisical
```
sheensantoscapadngan: replace-on-renewals: true

packages/cmd/agent.go
```diff
@@ -2295,16 +2314,80 @@ func failureRetryIntervalFor(certificate *AgentCertificateConfig) time.Duration
...
+
 func effectiveMaxFailureRetries(certificate *AgentCertificateConfig) int {
 	if certificate.Lifecycle.MaxFailureRetries > 0 {
 		return certificate.Lifecycle.MaxFailureRetries
 	}
 	return DEFAULT_MAX_FAILURE_RETRIES
 }
 
+func resolveCertificateFrom(httpClient *resty.Client, certificate *AgentCertificateConfig, anchorCertificateID string) (*api.RetrieveCertificateResponse, error) {
+	metadata, err := api.CallRetrieveCertificate(httpClient, anchorCertificateID)
```
sheensantoscapadngan: why is this called metadata?

packages/cmd/agent.go
```diff
@@ -2295,16 +2314,80 @@ func failureRetryIntervalFor(certificate *AgentCertificateConfig) time.Duration
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve certificate %s, the latest renewal of certificate %s: %v", latestID, anchorCertificateID, err)
+	}
+
+	return latest, nil
+}
+
+func effectiveCertificateStatus(metadata *api.RetrieveCertificateResponse) string {
+	status := metadata.Certificate.Status
+	if status == "active" && !metadata.Certificate.NotAfter.IsZero() && time.Now().After(metadata.Certificate.NotAfter) {
```
sheensantoscapadngan: let's use enums

e2e/agent/agent_helpers.go
```diff
@@ -588,8 +589,9 @@ type agentCertificateAttributes struct {
 }
 
 type agentCertificateLifecycle struct {
-	RenewBeforeExpiry   string `yaml:"renew-before-expiry"`
-	StatusCheckInterval string `yaml:"status-check-interval"`
+	RenewBeforeExpiry   string `yaml:"renew-before-expiry,omitempty"`
+	StatusCheckInterval string `yaml:"status-check-interval,omitempty"`
+	UseLatest           bool   `yaml:"use-latest,omitempty"`
```
sheensantoscapadngan: What hook do we trigger if you change the certificate ID manually and you have it so that it points to a completely different certificate altogether? It seems that we run on renewal instead of on replace. Feels like we should have a separate hook for that behavior

carlosmonastyrski: My thinking was that from the machine's point of view both cases are identical: the certificate on disk gets swapped for a different one and the service needs a reload. So `on-renewal` really means "the delivered certificate changed". Do you think that's fine, or would you rather have a separate hook?

sheensantoscapadngan (approved):

PR #375, opened 2026-08-26
fix(secrets): default secrets delete --type to shared to match secrets set

### LINKED ISSUE #7805: `infisical secrets delete` fails with 404 "Secret not found" for shared secrets because `--type` defaults to `personal`
### Product

Secret Manager

### Describe the bug

`infisical secrets delete <NAME>` always fails with HTTP 404 "Secret not found" for ordinary (shared) secrets, even ones created seconds earlier with `infisical secrets set`.

Cause: the delete command's `--type` flag defaults to `personal`, while `secrets set` defaults to `shared`. The backend then looks up a personal secret with that name, finds none, and returns 404 (`DELETE /api/v3/secrets/raw/<NAME>`). Passing `--type shared` explicitly makes the same delete succeed.

The mismatch is visible in `packages/cmd/secrets.go`: `secretsSetCmd` registers `--type` with default `util.SECRET_TYPE_SHARED`, `secretsDeleteCmd` registers it with default `"personal"`.

The error message ("Please make sure your secret path, workspace and environment name are all correct.") gives no hint that the secret type is the mismatched parameter.

### To Reproduce

1. Log in (`infisical login`, user auth) and connect a project (`infisical init`).
2. `infisical secrets set TEST_DELETE_ME=x --env dev`
3. `infisical secrets get TEST_DELETE_ME --env dev` — succeeds, shows type `shared`.
4. `infisical secrets delete TEST_DELETE_ME --env dev` — fails: `Response Code: 404 Not Found`, `Message: Secret not found` (e.g. request ID `req-eu-L2DaXAovylHn9p`).
5. `infisical secrets delete TEST_DELETE_ME --env dev --type shared` — succeeds.

Also reproduced with: `--path /`, explicit `--projectId`, plural name args, and running outside any `.infisical.json` directory — all 404 without `--type shared`.

### Expected behavior

Deleting a secret created by `infisical secrets set` should work without extra flags — the two commands should agree on the default secret type (or the delete should fall back to the shared secret when no personal one exists, or the error should name the type mismatch).

### Screenshots

N/A

### Deployment Type

Infisical Cloud

### Additional context

(Deployment is EU, https://eu.infisical.com)

- CLI v0.43.125 and v0.41.99 (linux amd64, GitHub release binaries) both reproduce identically — present at least since the CLI moved to the dedicated repo (commit `9c3717a4`), inherited from the monorepo.
- OS: Linux (WSL2, Ubuntu). Auth: interactive user login, file vault.
- `secrets set` (create and update), `secrets get`, `secrets` (list), `export`, `run`, and `secrets folders create/delete` all work against the same project/environment; only `secrets delete` is affected.
- Free tier, single project, default `/` path, no folders or approval policies involved.
- Docs mismatch: the CLI docs for `secrets set` state `--type` "defaults to shared", while the `secrets delete` docs do not document a `--type` flag at all — the `personal` default on delete is undocumented (https://infisical.com/docs/cli/commands/secrets).

rmdly:
Fixes Infisical/infisical#7805

`secrets set` creates a shared secret by default, but `secrets delete` looked for a personal one by default, so deleting a secret you had just created failed with a 404:

```
infisical secrets set TEST=x --env dev     # creates a shared secret
infisical secrets delete TEST --env dev    # 404 Secret not found
infisical secrets delete TEST --env dev --type shared   # works
```

The delete flag also hardcoded the string `"personal"` rather than using `util.SECRET_TYPE_PERSONAL`, and repeated the default in its usage text, which cobra already prints.

This aligns the two defaults on `util.SECRET_TYPE_SHARED`. Deleting a personal secret still works with an explicit `--type personal`.

Worth flagging: this is a behaviour change for anyone scripting `secrets delete` against personal secrets without passing `--type`. It seemed like the right call because `secrets set` already defaults to shared and the `personal` default on delete is not documented, but happy to switch to one of the other options in the issue if you would rather not change the default.

`TestSecretsTypeFlagDefaults` asserts both commands agree on the default so they cannot drift apart again.

One thing I noticed while adding it: nothing in CI runs `./packages/cmd`, and the package will not build under `go test` because of four pre-existing `non-constant format string` vet errors in `run.go`, so the existing unit tests there are not running either. I have left that alone here, but I am happy to open a separate PR fixing those if it would be useful.

### DIFF
--- a/packages/cmd/secrets.go
+++ b/packages/cmd/secrets.go
@@ -853,7 +853,7 @@ func init() {
 	secretsSetCmd.Flags().Bool("show-values", false, "reveal secret values in the output table instead of masking them")
 	util.AddOutputFlagsToCmd(secretsSetCmd, "The output to format the secrets in.")
 
-	secretsDeleteCmd.Flags().String("type", "personal", "the type of secret to delete: personal or shared  (default: personal)")
+	secretsDeleteCmd.Flags().String("type", util.SECRET_TYPE_SHARED, "the type of secret to delete: personal or shared")
 	secretsDeleteCmd.Flags().String("token", "", "Fetch secrets using service token or machine identity access token")
 	secretsDeleteCmd.Flags().String("projectId", "", "manually set the projectId to delete secrets from when using machine identity based auth")
 	secretsDeleteCmd.Flags().String("path", "/", "get secrets within a folder path")
--- a/packages/cmd/secrets_test.go
+++ b/packages/cmd/secrets_test.go
@@ -1,0 +1,28 @@
+package cmd
+
+import (
+	"testing"
+
+	"github.com/Infisical/infisical-merge/packages/util"
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSecretsTypeFlagDefaults(t *testing.T) {
+	tests := []struct {
+		name string
+		cmd  *cobra.Command
+	}{
+		{"secrets set", secretsSetCmd},
+		{"secrets delete", secretsDeleteCmd},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			flag := tc.cmd.Flags().Lookup("type")
+			require.NotNil(t, flag, "the type flag should be registered")
+			assert.Equal(t, util.SECRET_TYPE_SHARED, flag.DefValue)
+		})
+	}
+}


rmdly:
Gentle nudge on this one. CI has been green since 26 August and the CLA check is passing.

There is also the overlapping #381 flagged above, which does the same change without the regression test. Genuinely no preference from me on which lands, I would just rather it not sit indefinitely because of the duplication. If #381 is the one you would rather take, I am happy to send `TestSecretsTypeFlagDefaults` separately on top of it, and equally happy to change approach here if you would prefer the fallback or error-message option from the issue instead of flipping the default.

PR #377, opened 2026-08-27
feat(pam): default ssh access to interactive shell

lb-vn:
# Description 📣

Makes the SSH access CLI command default to an interactive shell, like web access. Before it would default to just starting a proxy.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/packages/cmd/pam.go
+++ b/packages/cmd/pam.go
@@ -28,13 +28,13 @@ var pamAccessCmd = &cobra.Command{
 	Use:   "access <path> [-- <command>]",
 	Short: "Launch a PAM session for the account at the given path",
 	Long: `Launch a PAM session for the account at the given path.
-The path format is: /folder/account-name (leading slash optional)
+The path format is: folder/account-name
 
 SSH accounts connect you straight to a shell on the target. Pass --proxy for a local
 proxy to point your own SSH, SCP or SFTP client at instead, or pass a command after
 '--' to run just that command and exit. Every other account type starts a local proxy
 or credential helper, which --proxy does not change.`,
-	Example: `  infisical pam access /production/postgres-main --duration 2h
+	Example: `  infisical pam access production/postgres-main --duration 2h
   infisical pam access servers/prod-bastion
   infisical pam access servers/prod-bastion -- systemctl status nginx
   infisical pam access servers/prod-bastion --proxy`,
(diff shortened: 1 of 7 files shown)


packages/cmd/pam.go
```diff
@@ -25,17 +25,32 @@ var pamCmd = &cobra.Command{
 }
 
 var pamAccessCmd = &cobra.Command{
-	Use:   "access <path>",
+	Use:   "access <path> [-- <command>]",
 	Short: "Launch a PAM session for the account at the given path",
 	Long: `Launch a PAM session for the account at the given path.
-The path format is: /folder/account-name (leading slash optional)`,
-	Example:               "infisical pam access /production/postgres-main --duration 2h",
+The path format is: /folder/account-name (leading slash optional)
```
bernie-g: I know this is unrelated to your PR but this help text is wrong since the leading slash is not even accepted. Can we update this

packages/cmd/pam.go
```diff
@@ -62,6 +77,17 @@ The path format is: /folder/account-name (leading slash optional)`,
...
 
+		proxy, err := cmd.Flags().GetBool("proxy")
+		if err != nil {
+			util.HandleError(err, "Unable to parse proxy flag")
+		}
+
+		if proxy && len(command) > 0 {
+			util.PrintErrorMessageAndExit("--proxy starts a local proxy for your own client, so it cannot also run a command on the target. Drop one of the two.")
+		}
+
```
bernie-g, line 88: ignore comment placement

Image

Hmm, I think it's a bit strange that if you use --port without --proxy in ssh it errors but --port with every other account type works fine. I feel like we should be more consistent somehow

packages/pam/local/access.go
```diff
@@ -702,6 +723,78 @@ func startSSHAccess(httpClient *resty.Client, response *api.PAMAccessResponse, p
 		return
 	}
 
+	if opts.Proxy {
+		startSSHProxy(httpClient, response, path, duration, username, opts.Port)
+		return
+	}
+
+	if opts.Port != 0 {
+		util.PrintErrorMessageAndExit("--port only applies to a local proxy. Add --proxy to start one, or drop --port to connect straight to a shell.")
```
bernie-g: Hm this gets checked after the pam session already starts so it can leave orphaned sessions. Was able to reproduce

Image

bernie-g (changes requested):

packages/cmd/pam.go
lb-vn: I just made --port be ignored when you reach the interactive terminal path

bernie-g (approved):

PR #378, opened 2026-08-27
feat: implement and refactor network request retries

scott-ray-wilson:
# Description 📣

Nothing in the CLI retried `429` or `504`. Resty's built-in default only retries transport errors, and it is silently *replaced* the moment a retry condition is added, so status codes were never inspected anywhere.

State before this PR:

| Surface | Retry behavior |
|---|---|
| `util.GetRestyClientWithCustomHeaders()`, **48 call sites** | none at all |
| 9 direct `resty.New()` sites (`agentproxy/*`, `pam/*`, `cmd/agent_proxy*.go`) | none, and they also skipped `INFISICAL_CUSTOM_HEADERS` |
| `util/helper.go`, `cmd/agent.go` | `SetRetryCount(10000)` with no condition |
| `cmd/login.go` | `SetRetryCount(5)`, same |

The `10000` is the tell: a connection refused retried ten thousand times while a `429` failed on the first attempt.

**Fix:** one policy in `packages/util/retry.go`, applied inside `GetRestyClientWithCustomHeaders()`. That single edit covers the 48 existing sites and every future `api.Call*`. The 9 direct constructions were converted, so there is now exactly one construction path, and all 5 ad-hoc overrides are gone.

**Policy**
- Retries `429`, `502`, `503`, `504`, plus typed transport errors (`net.Error`, `ECONNRESET`/`ECONNREFUSED`/`EPIPE`/`EHOSTUNREACH`/`ENETUNREACH`/`ETIMEDOUT`, `io.EOF`). Type and errno checks, not substring matching on error text.
- Honors `Retry-After` in both RFC 9110 forms, capped at `MaxDelay`. Nothing read that header before.
- **POST retries only on `429`.** A `502`/`503`/`504` can mean the server did process the write and only the response was lost, so replaying a POST risks double-applying it, for instance minting a second dynamic secret lease. `429` is safe because the server states it rejected the request outright.
- Does not retry `4xx`, bare `500`, TLS trust failures, or `context.Canceled`/`DeadlineExceeded`. Allow-list, so anything unrecognised surfaces immediately.
- Three named policies, all honoring `INFISICAL_RETRY_*`: `DefaultRetryPolicy` (3 retries / 10s), `AgentRetryPolicy` (30 / 30s, for the agent's token lifecycle), `BestEffortRetryPolicy` (1 / 1s, for agent-proxy usage reporting, which also runs on the shutdown path).

`TestNoDirectRestyConstruction` walks `packages/` and fails on any new `resty.New()`. Without it the choke point silently decays: a direct construction compiles, runs, and looks fine in review, it just has no retries.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

### Behavior changes worth a look

- Agent token refresh: effectively-infinite retries capped at 30 (`AgentRetryPolicy`).
- `cmd/login.go`: 5 retries becomes 3, but now covers `429`/`504` where it previously covered neither.
- Malformed `INFISICAL_RETRY_*` warns and falls back to the default instead of `os.Exit(1)`, since this now runs while building a client for any command. The agent's stricter SDK-side validation is unchanged.
- The 9 converted sites now honor `INFISICAL_CUSTOM_HEADERS`. Latent fix, but a behavior change.

### Known gap

17 sites build `infisicalSdk.NewInfisicalClient` (`ssh`, `dynamic-secrets`, `gateway`, `pam`, agent templating). The SDK's own retry condition never inspects status codes either, so `429`/`504` there still fail on the first attempt. Closing that needs a `go-sdk` change, since its opt-in `RetryRequestsConfig` retries any `IsError()` including `401` and `404`. Out of scope here. The high-traffic paths are covered: `infisical run`, `secrets get`, and `export` all go through `packages/util/secrets.go`, which uses the shared constructor.

Also still outside resty entirely: `packages/pam/session/chunk_uploader.go:458` uses `http.DefaultClient`.

# Tests 🛠️

Unit tests in `packages/util/retry_test.go` cover the status-code matrix, POST-vs-GET method safety, `Retry-After` parsing and capping, transport-error classification (including that TLS trust failures are *not* retried), env overrides, and the choke-point guard.

```sh
go build ./...
go test ./packages/... -count=1 -vet=off
go vet ./packages/util/ ./packages/agentproxy/ ./packages/pam/...
```

The guard test was verified to actually fail on a planted `resty.New()`, not just pass vacuously.

## Manual verification

Drove the built binary against a mock API, counting requests server-side rather than trusting the CLI's own logs.

```sh
cat > /tmp/mock.py <<'PY'
import json, os, time
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
SEQ = [int(x) for x in os.environ.get("SEQUENCE", "200").split(",")]  # last value repeats
RA, PORT = os.environ.get("RETRY_AFTER", ""), int(os.environ.get("PORT", "8899"))
hits, start = [], time.monotonic()
class H(BaseHTTPRequestHandler):
    protocol_version = "HTTP/1.1"
    def _h(self):
        n = int(self.headers.get("Content-Length") or 0)
        if n: self.rfile.read(n)   # must drain, or the next keep-alive request is mangled
        i = len(hits); s = SEQ[i] if i < len(SEQ) else SEQ[-1]; hits.append(s)
        print(f"HIT #{i+1} t=+{time.monotonic()-start:5.2f}s {self.command} -> {s}", flush=True)
        b = json.dumps({"accessToken": "renewed", "accessTokenTTL": 3600}).encode()
        self.send_response(s)
        if RA and s in (429, 503): self.send_header("Retry-After", RA)
        self.send_header("Content-Type", "application/json"); self.send_header("Content-Length", str(len(b)))
        self.end_headers(); self.wfile.write(b)
    do_GET = do_POST = do_PUT = do_DELETE = do_PATCH = _h
    def log_message(self, *a): pass
ThreadingHTTPServer(("127.0.0.1", PORT), H).serve_forever()
PY

go build -o /tmp/infisical .

# 1. POST + 429 twice then 200 -> retries, then succeeds
SEQUENCE=429,429,200 PORT=8901 python3 /tmp/mock.py &
/tmp/infisical token renew ua.mock --domain http://127.0.0.1:8901 --log-level debug

# 2. POST + 504 -> must NOT retry (unsafe to replay a write)
SEQUENCE=504 PORT=8902 python3 /tmp/mock.py &
/tmp/infisical token renew ua.mock --domain http://127.0.0.1:8902 --log-level debug

# 3. Retry-After honored (gaps ~2s, not the 500ms base delay)
SEQUENCE=429,429,200 RETRY_AFTER=2 PORT=8903 python3 /tmp/mock.py &
/tmp/infisical token renew ua.mock --domain http://127.0.0.1:8903 --log-level debug

# 4. GET + 504 -> must retry (safe to replay a read); this is the `infisical run` path
SEQUENCE=504 PORT=8904 python3 /tmp/mock.py &
INFISICAL_TOKEN=ua.mock /tmp/infisical secrets --projectId p --env dev --domain http://127.0.0.1:8904 --log-level debug

# 5. Env overrides
SEQUENCE=504 PORT=8905 python3 /tmp/mock.py &
INFISICAL_TOKEN=ua.mock INFISICAL_RETRY_MAX_RETRIES=6 INFISICAL_RETRY_BASE_DELAY=100ms \
  /tmp/infisical secrets --projectId p --env dev --domain http://127.0.0.1:8905 --log-level debug

# 6. Transport error: bind and release a port so nothing is listening
/tmp/infisical token renew ua.mock --domain http://127.0.0.1:59889 --log-level debug
```

Results:

| # | Scenario | Expected | Observed |
|---|---|---|---|
| 1 | POST, 429 ×2 then 200 | retry then succeed | 3 hits at 0.05 / 0.55 / 1.38s, succeeded |
| 2 | POST, 504 | no retry | 1 hit |
| 3 | POST, 429 + `Retry-After: 2` | ~2s gaps | 2.00s and 2.01s |
| 4 | GET, 504 | 4 attempts | 4 hits, 0.15 / 0.65 / 1.50 / 3.20s |
| 5 | 6 retries, 100ms base | 7 attempts, ~100ms first gap | 7 hits, first gap 0.11s |
| 6 | connection refused | 4 attempts | 4 attempts, then "retries are exhausted" |
| - | POST, 401 | no retry | 1 hit |

**Not run:** the `e2e/` suite. It builds and vets clean (`cd e2e && go build ./... && go vet ./...`), but running it needs `INFISICAL_BACKEND_DIR` and an `e2e/.env`. Nothing was tested against a real Infisical instance.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

This PR adds/unifies request retry attempts

## Type ✨

- [ ] Bug fix
- [x] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/packages/util/helper.go
+++ b/packages/util/helper.go
@@ -329,7 +329,10 @@ func GetInfisicalToken(cmd *cobra.Command) (token *models.TokenDetails, err erro
 }
 
 func UniversalAuthLogin(clientId string, clientSecret string) (api.UniversalAuthLoginResponse, error) {
-	httpClient, err := GetRestyClientWithCustomHeaders()
+	policy := DefaultRetryPolicy()
+	policy.ReplaySafe = true // a replayed login only mints another TTL-bound token
+
+	httpClient, err := GetRestyClientWithPolicy(policy)
 	if err != nil {
 		return api.UniversalAuthLoginResponse{}, err
 	}
@@ -343,8 +346,10 @@ func UniversalAuthLogin(clientId string, clientSecret string) (api.UniversalAuth
 }
 
 func RenewMachineIdentityAccessToken(accessToken string) (string, error) {
+	policy := DefaultRetryPolicy()
+	policy.ReplaySafe = true // renewal extends the presented token, so a replay cannot double-apply
 
-	httpClient, err := GetRestyClientWithCustomHeaders()
+	httpClient, err := GetRestyClientWithPolicy(policy)
 	if err != nil {
 		return "", err
 	}
--- a/packages/util/helper.go
+++ b/packages/util/helper.go
@@ -334,10 +334,6 @@ func UniversalAuthLogin(clientId string, clientSecret string) (api.UniversalAuth
 		return api.UniversalAuthLoginResponse{}, err
 	}
 
-	httpClient.SetRetryCount(10000).
-		SetRetryMaxWaitTime(20 * time.Second).
-		SetRetryWaitTime(5 * time.Second)
-
 	tokenResponse, err := api.CallUniversalAuthLogin(httpClient, api.UniversalAuthLoginRequest{ClientId: clientId, ClientSecret: clientSecret})
 	if err != nil {
 		return api.UniversalAuthLoginResponse{}, err
@@ -352,10 +348,6 @@ func RenewMachineIdentityAccessToken(accessToken string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-
-	httpClient.SetRetryCount(10000).
-		SetRetryMaxWaitTime(20 * time.Second).
-		SetRetryWaitTime(5 * time.Second)
 
 	request := api.UniversalAuthRefreshRequest{
 		AccessToken: accessToken,
--- a/packages/util/helper.go
+++ b/packages/util/helper.go
@@ -329,10 +329,7 @@ func GetInfisicalToken(cmd *cobra.Command) (token *models.TokenDetails, err erro
 }
 
 func UniversalAuthLogin(clientId string, clientSecret string) (api.UniversalAuthLoginResponse, error) {
-	policy := DefaultRetryPolicy()
-	policy.ReplaySafe = true // a replayed login only mints another TTL-bound token
-
-	httpClient, err := GetRestyClientWithPolicy(policy)
+	httpClient, err := GetRestyClientWithCustomHeaders()
 	if err != nil {
 		return api.UniversalAuthLoginResponse{}, err
 	}
--- a/packages/util/retry.go
+++ b/packages/util/retry.go
@@ -1,0 +1,331 @@
+package util
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"io"
+	"net"
+	"net/http"
+	"os"
+	"strconv"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/go-resty/resty/v2"
+	"github.com/rs/zerolog/log"
+)
+
+// Retries live here so request sites don't configure them. Every resty client used to talk to the
+// Infisical API is built by GetRestyClientWithCustomHeaders, which applies DefaultRetryPolicy, so a
+// new api.Call* inherits retries without doing anything. Long-running commands that should ride out
+// an outage instead of exiting use GetRestyClientWithPolicy(AgentRetryPolicy()).
+
+const (
+	defaultRetryMaxRetries = 3
+	defaultRetryBaseDelay  = 500 * time.Millisecond
+	defaultRetryMaxDelay   = 10 * time.Second
+
+	// Long-running commands (agent, gateway, relay) keep trying for roughly a quarter hour so a
+	// brief upstream outage doesn't take the process down with it.
+	agentRetryMaxRetries = 30
+	agentRetryMaxDelay   = 30 * time.Second
+)
+
+// retryableStatusCodes are the responses worth repeating. Deliberately narrow: a 401 or 404 does not
+// improve on the third attempt, and retrying it just multiplies the cost of a bad token or a typo.
+var retryableStatusCodes = map[int]bool{
+	http.StatusTooManyRequests:    true, // 429
+	http.StatusBadGateway:         true, // 502
+	http.StatusServiceUnavailable: true, // 503
+	http.StatusGatewayTimeout:     true, // 504
+}
+
+// RetryPolicy bounds how a resty client retries transient API failures. Time spent sleeping between
+// attempts is at most MaxRetries * MaxDelay, so those two fields together decide how long a command
+// keeps trying before it gives up.
+type RetryPolicy struct {
+	// MaxRetries counts attempts after the first, matching resty's SetRetryCount. Zero disables
+	// retries.
+	MaxRetries int
+	BaseDelay  time.Duration
+	MaxDelay   time.Duration
+}
+
+// DefaultRetryPolicy suits one-shot commands, where a user or script is waiting on the result and
+// would rather see the error than sit through a long backoff.
+func DefaultRetryPolicy() RetryPolicy {
+	return RetryPolicy{
+		MaxRetries: defaultRetryMaxRetries,
+		BaseDelay:  defaultRetryBaseDelay,
+		MaxDelay:   defaultRetryMaxDelay,
+	}.withEnvOverrides()
+}
+
+// BestEffortRetryPolicy suits calls whose failure is tolerable and whose latency is not, such as
+// usage reporting that also runs on the shutdown path. It absorbs a single blip without making the
+// user wait on a full backoff for a result nobody reads.
+func BestEffortRetryPolicy() RetryPolicy {
+	return RetryPolicy{
+		MaxRetries: 1,
+		BaseDelay:  200 * time.Millisecond,
+		MaxDelay:   time.Second,
+	}.withEnvOverrides()
+}
+
+// AgentRetryPolicy suits processes expected to outlive a transient outage rather than exit on one.
+func AgentRetryPolicy() RetryPolicy {
+	return RetryPolicy{
+		MaxRetries: agentRetryMaxRetries,
+		BaseDelay:  defaultRetryBaseDelay,
+		MaxDelay:   agentRetryMaxDelay,
+	}.withEnvOverrides()
+}
+
+// withEnvOverrides applies the INFISICAL_RETRY_* variables on top of whichever defaults were chosen.
+// A malformed value is reported and skipped rather than fatal, since this runs while building a
+// client for a command that would otherwise work fine.
+func (p RetryPolicy) withEnvOverrides() RetryPolicy {
+	if raw := os.Getenv(INFISICAL_RETRY_BASE_DELAY_NAME); raw != "" {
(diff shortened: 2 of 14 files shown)


scott-ray-wilson:
@greptile review

packages/util/retry.go
```diff
@@ -0,0 +1,376 @@
...
+
+	var certInvalidErr x509.CertificateInvalidError
+	if errors.As(err, &certInvalidErr) {
+		return false
+	}
+
+	// Covers dial timeouts, DNS failures and refused connections: http.Client wraps transport
+	// failures in *url.Error, which satisfies net.Error.
+	var netErr net.Error
+	if errors.As(err, &netErr) {
```
Thiago-AS, line 207: I think http.Client.Do wraps every transport failure in *url.Error. So the errors.As(err, &netErr) match's everything. So I don't think it goes past that point.

packages/util/retry.go
```diff
@@ -0,0 +1,376 @@
...
+	// Resolution never produced an address, so nothing was sent.
+	var dnsErr *net.DNSError
+	if errors.As(err, &dnsErr) {
+		return true
+	}
+
+	// Op is "dial" only while establishing the connection. A request already in flight reports
+	// "read" or "write" instead.
+	var opErr *net.OpError
+	if errors.As(err, &opErr) && opErr.Op == "dial" {
```
Thiago-AS: Same here, refused connections arrive as *net.OpError and are already caught by this check so line 236 never fires.

packages/util/retry.go
```diff
@@ -0,0 +1,376 @@
...
+	switch method {
+	case http.MethodGet, http.MethodHead, http.MethodOptions, http.MethodTrace,
+		http.MethodPut, http.MethodDelete:
+		return true
+	default: // POST, PATCH, anything unrecognised
+		return false
+	}
+}
+
+// requestNeverReachedServer reports whether a transport error proves the request was never delivered,
```
Thiago-AS: I think there’s a bigger point here worth calling out. The “never reached the server” logic only covers DNS, dial, and refused connection cases. The more common case mentioned below, a keep-alive connection getting reset (ECONNRESET/EOF), can happen after the POST was already written, so it’s ambiguous and this code won’t retry it.

That means token refresh POSTs can go from being retried by the old `SetRetryCount(10000)` behavior to only 1 attempt here. For example, POST + 503 = 1 attempt, while GET + 503 = 4 attempts.

I think we should make an explicit call on whether token refresh is considered replay safe, rather than letting the HTTP method gating decide that implicitly. If it is replay-safe, we probably want that represented explicitly in the retry policy.

packages/util/retry.go
```diff
@@ -0,0 +1,376 @@
...
+	}
+
+	wait, ok := parseRetryAfter(res.Header().Get("Retry-After"))
+	if !ok {
+		return 0
+	}
+
+	// The server told us exactly how long to wait, so prefer it over our own guess. Still cap it:
+	// a misconfigured or hostile Retry-After shouldn't be able to park the CLI indefinitely.
+	if policy.MaxDelay > 0 && wait > policy.MaxDelay {
```
Thiago-AS: A 429 carrying `Retry-After: 60` blocks 10s instead of failing fast, and with MaxRetries: 3 a rate-limited one-shot command can block ~30s while repeatedly hitting a server that asked for 60. Maybe we should return an error instead of retrying before?

packages/util/retry.go
```diff
@@ -0,0 +1,376 @@
...
+// Infisical API is built by GetRestyClientWithCustomHeaders, which applies DefaultRetryPolicy, so a
+// new api.Call* inherits retries without doing anything. Long-running commands that should ride out
+// an outage instead of exiting use GetRestyClientWithPolicy(AgentRetryPolicy()).
+
+const (
+	defaultRetryMaxRetries = 3
+	defaultRetryBaseDelay  = 500 * time.Millisecond
+	defaultRetryMaxDelay   = 10 * time.Second
+
+	// Long-running commands (agent, gateway, relay) keep trying for roughly a quarter hour so a
```
Thiago-AS: Checked and I don't think gateway and relay are using AgentRetryPolicy.

packages/util/retry.go
```diff
@@ -0,0 +1,376 @@
...
+		BaseDelay:  defaultRetryBaseDelay,
+		MaxDelay:   agentRetryMaxDelay,
+	}.withEnvOverrides()
+}
+
+// withEnvOverrides applies the INFISICAL_RETRY_* variables on top of whichever defaults were chosen.
+// A malformed value is reported and skipped rather than fatal, since this runs while building a
+// client for a command that would otherwise work fine.
+func (p RetryPolicy) withEnvOverrides() RetryPolicy {
+	if raw := os.Getenv(INFISICAL_RETRY_BASE_DELAY_NAME); raw != "" {
```
Thiago-AS, line 78: retry:
  max-retries: ...
  base-delay: ...
  max-delay: ...

The vars are now environment for every single http call, not command scoped. A Helm chart, Docker Compose file, or k8s manifest that sets INFISICAL_RETRY_MAX_RETRIES to tune the agent now also changes every other CLI invocation sharing that environment, is that intended? We use to only use that for agents before, so lets say someone had a huge max-retry or timeout because an agent, that is now replied to all http calls. 

Just flagging.

Thiago-AS (changes requested):
Maybe we should tone down the comments lol. There are a lot of them, and they can get stale pretty easily. Most of them are just stating the obvious.

Claude, go over the comments that were added and leave only what’s necessary. Remove anything obvious or self-documented by the code. Keep the comments concise. No em dashes XD

packages/util/retry.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	if p.MaxDelay > 0 && p.BaseDelay > p.MaxDelay {
+		p.BaseDelay = p.MaxDelay
+	}
+
+	return p
+}
+
+func applyRetryPolicy(httpClient *resty.Client, policy RetryPolicy) *resty.Client {
```
Thiago-AS, line 109: Now that every client built through GetRestyClientWithPolicy has retries, each failure emits unstructured logs to os.Stderr through resty's own logger, bypassing structure logs:

```
2026/08/28 14:53:41.746923 WARN RESTY Get "http://127.0.0.1:55497": dial tcp ...: connection refused, Attempt 1
2026/08/28 14:53:41.750776 ERROR RESTY Get "http://127.0.0.1:55497": dial tcp ...: connection refused
```

packages/util/helper.go
```diff
@@ -329,15 +329,14 @@ func GetInfisicalToken(cmd *cobra.Command) (token *models.TokenDetails, err erro
 }
 
 func UniversalAuthLogin(clientId string, clientSecret string) (api.UniversalAuthLoginResponse, error) {
-	httpClient, err := GetRestyClientWithCustomHeaders()
+	policy := DefaultRetryPolicy()
+	policy.ReplaySafe = true // a replayed login only mints another TTL-bound token
```
Thiago-AS: I think this is bad because UA can have limit usage. I think we should drop the ReplaySafe from this one,  login still replays on pre-delivery failures (connection refused, DNS), which is the common transient case.

Thiago-AS (changes requested):

Thiago-AS (approved):

PR #379, opened 2026-08-27
feat(pam): heartbeats

bernie-g:
# Description 📣

The test-connection and WinRM handlers returned the same 502 for a refused credential and an unreachable target, so the control plane could not tell them apart. They now classify the failure at the source and return it as `auth`, `transport`, or `unknown`.

Also routes MSSQL Windows-auth logins through the session proxy's handshake, which already speaks NTLM and Kerberos.

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Unit tests for the MSSQL auth routing, plus a live NTLM check (env-gated, skipped in CI) verified against a real SQL Server 2022 instance: correct password authenticates, wrong password is rejected.

```sh
go test ./packages/gateway-v2/ ./packages/pam/handlers/mssql/
```


packages/gateway-v2/test_connection_handler.go
```diff
@@ -205,6 +211,23 @@ func openSQLTestDB(host string, port int, params sqlTestParams) (*sql.DB, error)
 
 // doSQLConnectionTest authenticates against the target SQL server and runs a trivial query
 func doSQLConnectionTest(ctx context.Context, host string, port int, params sqlTestParams) error {
+	// Windows-auth SQL Server logins have no SQL-managed password, so the database/sql driver path can't carry them.
+	// The session proxy already speaks NTLM and Kerberos against MSSQL, so the check reuses that handshake.
+	if params.Dialect == "mssql" && (params.AuthMethod == "ntlm" || params.AuthMethod == "kerberos") {
+		return mssqlhandler.VerifyCredential(mssqlhandler.MssqlProxyConfig{
```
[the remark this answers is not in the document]
bernie-g: Fixed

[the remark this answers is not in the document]
bernie-g: Fixed

packages/gateway-v2/test_connection_failure_kind.go
```diff
@@ -0,0 +1,139 @@
...
+			return failureKindUnknown
+		}
+	}
+
+	var msErr mssql.Error
+	if errors.As(err, &msErr) {
+		// 18456 login failed, 18452 untrusted domain, 4060 cannot open database for this login.
+		switch msErr.Number {
+		case 4060, 18452, 18456:
+			return failureKindAuth
```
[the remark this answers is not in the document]
bernie-g: I think this is fine

packages/gateway-v2/test_connection_handler.go
```diff
@@ -205,6 +211,23 @@ func openSQLTestDB(host string, port int, params sqlTestParams) (*sql.DB, error)
...
+			InjectUsername: params.Username,
+			InjectPassword: params.Password,
+			InjectDatabase: params.Database,
+			InjectDomain:   params.Domain,
+			InjectRealm:    params.Realm,
+			InjectKDCAddr:  params.KdcAddress,
+			InjectSPN:      params.Spn,
+			AuthMethod:     params.AuthMethod,
+			EnableTLS:      params.SslEnabled,
+		})
```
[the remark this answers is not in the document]
bernie-g: Fixed

packages/gateway-v2/test_connection_failure_kind.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	return classifyTestConnFailureByMessage(err.Error())
+}
+
+// Drivers without typed errors only report a string.
+var authFailureSubstrings = []string{
+	"unable to authenticate",      // golang.org/x/crypto/ssh
+	"no supported methods remain", // golang.org/x/crypto/ssh
+	"ssh: handshake failed",       // golang.org/x/crypto/ssh
```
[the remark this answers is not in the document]
bernie-g: Fixed

packages/gateway-v2/test_connection_handler.go
```diff
@@ -383,12 +422,12 @@ func doKubernetesConnectionTest(ctx context.Context, host string, port int, para
...
+		return connectFailure(err)
 	}
 	defer resp.Body.Close()
 
 	if resp.StatusCode == http.StatusUnauthorized {
-		return fmt.Errorf("kubernetes API rejected the credentials (HTTP %d)", resp.StatusCode)
+		return authFailure(fmt.Errorf("kubernetes API rejected the credentials (HTTP %d)", resp.StatusCode))
 	}
 	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
 		return fmt.Errorf("kubernetes API returned HTTP %d", resp.StatusCode)
```
lb-vn: A 5xx from the api server falls through untagged, and the platform reads untagged as a rejected credential. connectFailure here? Same at line 512

packages/gateway-v2/test_connection_handler.go
```diff
@@ -205,14 +211,42 @@ func openSQLTestDB(host string, port int, params sqlTestParams) (*sql.DB, error)
...
+			InjectRealm:    params.Realm,
+			InjectKDCAddr:  params.KdcAddress,
+			InjectSPN:      params.Spn,
+			AuthMethod:     params.AuthMethod,
+			EnableTLS:      params.SslEnabled,
+			TLSConfig:      tlsConfig,
+		}))
+	}
+
 	db, err := openSQLTestDB(host, port, params)
```
lb-vn, line 242: Claude:

openSQLTestDB only fails on dsn and tls config, but its error goes back untagged and gets read as a rejected credential. same shape with mongo.Connect and the buildTestTLSConfig call in the ntlm branch

packages/pam/handlers/mssql/proxy.go
```diff
@@ -654,3 +659,27 @@ func (p *MssqlProxy) proxyToClient(server, client net.Conn, errCh chan error) {
...
+			return fmt.Errorf("set handshake deadline: %w", err)
+		}
+	}
+
+	proxy := NewMssqlProxy(config)
+	authedConn, _, err := proxy.authenticateOverConn(serverConn)
+	if err != nil {
+		return err
+	}
+	return authedConn.Close()
```
lb-vn: Claude:
Returning the result of Close() may imply that the probe didn't work if an error happens. At this point we can just return success

lb-vn (changes requested):

packages/gateway-v2/test_connection_handler.go
bernie-g: Fixed

packages/gateway-v2/test_connection_handler.go
bernie-g: Fixed

packages/pam/handlers/mssql/proxy.go
bernie-g: Fixed

lb-vn (approved):

PR #383, opened 2026-09-03
chore(goreleaser): change release mode to keep-existing in configurat…

victorvhs017:
# Description 📣

The release pipeline runs GoReleaser three times against the same tag — `goreleaser` (linux), `goreleaser-darwin` (`.goreleaser-darwin.yaml`), and `goreleaser-windows` (`.goreleaser-windows.yaml`). All three had `release.mode: append`, and none of them define a `changelog:` section, so each job generated the same default git changelog and appended it to the GitHub release body. The result was the release notes repeated up to three times per release, and more on every re-run of a job.

This switches all three configs to `release.mode: keep-existing`. The first job to reach the release step creates the release with its notes; the other two upload their artifacts and leave the existing body untouched. `replace_existing_artifacts: true` is unchanged, so re-runs are still idempotent for assets (a 422 `already_exists` deletes the conflicting asset and retries).

# Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/.goreleaser-darwin.yaml
+++ b/.goreleaser-darwin.yaml
@@ -47,7 +47,7 @@ archives:
 
 release:
   # replace_existing_artifacts so re-runs don't 422 on already-uploaded assets.
-  mode: append
+  mode: keep-existing
   replace_existing_artifacts: true
 
 checksum:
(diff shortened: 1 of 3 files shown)


Vligai (approved):
LGTM! We should review if we ever move to distribution specific releases.

PR #387, opened 2026-09-04
cli: support multiple --path values in export command

### LINKED ISSUE #900: Specify multiple paths in the CLI

mateuscmtropical:
This implements #900 (specify multiple paths in the CLI).

`run` already supported multiple `--path` flags. `export` only took one. This brings `export` up to the same behavior:

- `--path` is now a `StringArray`, matching `run`'s existing flag
- reuses `run`'s `fetchSecrets` to fetch secrets from each path
- merges duplicate keys across paths (last path wins)
- tag filtering now happens once, client-side, after the merge, instead of being applied per-path on the server — otherwise a tagged value from an earlier path could survive over a later path's untagged override
- fixed a related bug in error handling: wrapped API errors weren't unwrapped before the pretty-print check, so failures showed a generic error instead of the formatted one

Tested manually against a real project (multiple folders, with and without tag filtering) and with `go test ./packages/cmd/... -count=2`.

Closes Infisical/infisical#900

Student contribution for a free software course. Feedback welcome.

### DIFF
--- a/packages/cmd/export.go
+++ b/packages/cmd/export.go
@@ -83,7 +83,7 @@ var exportCmd = &cobra.Command{
 			util.HandleError(err, "Unable to parse flag")
 		}
 
-		secretsPath, err := cmd.Flags().GetString("path")
+		secretsPaths, err := cmd.Flags().GetStringArray("path")
 		if err != nil {
 			util.HandleError(err, "Unable to parse flag")
 		}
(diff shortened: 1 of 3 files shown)


mateuscmtropical:
Closing in favor of a fresh PR from a clean branch.

mateuscmtropical:
Superseded by #388.

PR #388, opened 2026-09-04
feat: support multiple --path values in export command

mateuscmtropical:
# Description 📣

Implements #900 (specify multiple paths in the CLI).

`run` already supported multiple `--path` flags. `export` only took one. This brings `export` up to the same behavior:

- `--path` is now a `StringArray`, matching `run`'s existing flag
- reuses `run`'s `fetchSecrets` to fetch secrets from each path
- merges duplicate keys across paths (last path wins)
- tag filtering now happens once, client-side, after the merge, instead of being applied per-path on the server — otherwise a tagged value from an earlier path could survive over a later path's untagged override
- fixed a related bug in error handling: wrapped API errors weren't unwrapped before the pretty-print check, so failures showed a generic error instead of the formatted one

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Ran `go test ./packages/cmd/... -vet=off -count=2` (all passing).

Also tested manually against a real Infisical Cloud project: created two folders with overlapping and non-overlapping secret keys, some tagged and some not, and confirmed:

```sh
infisical export --env=dev --path=/foo --path=/bar
infisical export --env=dev --path=/foo --path=/bar --tags=backend
```

both return the expected merged/filtered result (last path wins on key collisions, tag filtering applied after merge).

---

### DIFF
--- a/packages/cmd/export.go
+++ b/packages/cmd/export.go
@@ -83,7 +83,7 @@ var exportCmd = &cobra.Command{
 			util.HandleError(err, "Unable to parse flag")
 		}
 
-		secretsPath, err := cmd.Flags().GetString("path")
+		secretsPaths, err := cmd.Flags().GetStringArray("path")
 		if err != nil {
 			util.HandleError(err, "Unable to parse flag")
 		}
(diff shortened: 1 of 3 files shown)


mateuscmtropical:
Hi @varonix0 @carlosmonastyrski, could you take a look at this when you have a moment? Happy to adjust anything.

PR #390, opened 2026-09-07
improvement(login): treat INFISICAL_DOMAIN env as explicit instance selection for cloud URLs

victorvhs017:
# Description 📣

Setting `INFISICAL_DOMAIN` or legacy `INFISICAL_API_URL` should be treated as an explicit instance choice—the same as passing `--domain`. Before this change, user login only honored those env vars for non-default hosts; US Cloud (`https://app.infisical.com`) and EU Cloud (`https://eu.infisical.com`) from env were ignored and the interactive hosting picker still appeared.

Login now treats `--domain`, `INFISICAL_DOMAIN`, and `INFISICAL_API_URL` as explicit domain sources. When any of them is set (including US/EU cloud URLs), `usePresetDomain` skips the hosting picker and uses that instance. The INFO line names the source (`--domain flag`, `INFISICAL_DOMAIN environment variable`, or `INFISICAL_API_URL environment variable`). `--domain` still wins when both flag and env are set (via existing `resolveDomain` precedence).

Supporting changes: `GetEnvDomainSource()` returns the domain and which env var supplied it; `AppendAPIEndpoint` trims trailing slashes before appending `/api`; domain normalization is extracted into small helpers so origins ending in `/api` are not over-stripped.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

1. With no flag or env set, run `infisical login` and confirm the hosting picker appears.
2. Run `INFISICAL_DOMAIN=https://eu.infisical.com infisical login` and confirm:
   - INFO mentions `INFISICAL_DOMAIN environment variable`
   - No "Select your hosting option" prompt
   - Browser/callback targets `eu.infisical.com`
3. Repeat with `INFISICAL_DOMAIN=https://app.infisical.com` for US Cloud.
4. Run `infisical login --domain https://eu.infisical.com` with env unset; INFO should say `--domain flag`.
5. With both set (`INFISICAL_DOMAIN=https://app.infisical.com` and `--domain https://eu.infisical.com`), confirm the flag wins and INFO says `--domain flag`.
6. Run `INFISICAL_API_URL=https://eu.infisical.com infisical login` (no `INFISICAL_DOMAIN`) and confirm legacy env behaves like explicit domain selection.
7. Run `INFISICAL_DOMAIN=$API_URL infisical login --method=universal-auth --plain --silent` against a local/self-hosted instance and confirm login succeeds without the picker.

---

### DIFF
--- a/packages/cmd/login.go
+++ b/packages/cmd/login.go
@@ -153,8 +153,9 @@ var loginCmd = &cobra.Command{
 			}
 
 			domainFlagExplicitlySet := cmd.Flags().Changed("domain")
+			_, domainEnvName, domainFromEnv := util.GetEnvDomainSource()
 			shouldPrintInfo := !silentMode && !plainOutput
-			usePresetDomain, err := usePresetDomain(presetDomain, domainFlagExplicitlySet, shouldPrintInfo)
+			usePresetDomain, err := usePresetDomain(presetDomain, domainFlagExplicitlySet, domainFromEnv, domainEnvName, shouldPrintInfo)
 
 			if err != nil {
 				util.HandleError(err)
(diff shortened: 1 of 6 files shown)


packages/cmd/login.go
```diff
@@ -452,21 +453,40 @@ func DomainOverridePrompt() (bool, error) {
...
-	preconfiguredUrl := strings.TrimSuffix(presetDomain, "/api")
-
-	// If the domain flag was explicitly set by the user, use it directly (even for US/EU cloud URLs)
-	// Otherwise, only use the preset domain if it's not a default cloud URL
-	shouldUsePresetDomain := preconfiguredUrl != "" && (domainFlagExplicitlySet || (preconfiguredUrl != util.INFISICAL_DEFAULT_US_URL && preconfiguredUrl != util.INFISICAL_DEFAULT_EU_URL))
-
-	if shouldUsePresetDomain {
-		parsedDomain := strings.TrimSuffix(strings.Trim(preconfiguredUrl, "/"), "/api")
+	parsedDomain := normalizePresetDomain(presetDomain)
+	domainExplicitlyProvided := domainFlagExplicitlySet || domainEnvName != ""
```
Thiago-AS: Previously, a domain from `.infisical.json` was applied silently. Now, after domain normalization, it’s treated as non-explicit, which triggers the domain picker. For EU or US cloud URLs, picking the default can then incorrectly send the user to the US domain.

The normalization change itself looks correct. The issue is that the source from `.infisical.json` is no longer recognized as explicit.

- .infisical.json containing {"domain": "https://eu.infisical.com/api/"}  no flag, no env, then infisical login:
    - before: EU applied silently, no picker
    - after: falls through to askForDomain(), and picking the default lands the user on US

Thiago-AS (approved):

PR #391, opened 2026-09-08
feat(pam): add Oracle credential rotation over the gateway

carlosmonastyrski:
# Description 📣

The Gateway can now change an Oracle user's password on Infisical's behalf, which is what makes Oracle credential rotation and health checks work end to end.

It also handles the awkward parts of Oracle specifically: databases that enforce a password policy, which reject a password change unless the old one is supplied, and accounts reached over an encrypted connection, which Infisical could not check on its own. Connecting to an Oracle account through the CLI now tells you to type `password` in your client's password field, Oracle is the one database that refuses to log in without something in that box, and the Gateway swaps in the real credential for you.

Error messages from failed Oracle connections no longer echo the connection string back, so a password can't end up in a status message.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -45,6 +45,7 @@ require (
 	github.com/quic-go/quic-go v0.59.1
 	github.com/rs/cors v1.11.0
 	github.com/rs/zerolog v1.35.1
+	github.com/sijms/go-ora/v2 v2.9.0
 	github.com/smallnest/resp3 v0.0.0-20251228151914-4f2fa7427e69
 	github.com/spf13/cobra v1.6.1
 	github.com/spf13/viper v1.8.1
@@ -104,7 +105,7 @@ require (
 	github.com/dustin/go-humanize v1.0.0 // indirect
 	github.com/dvsekhvalnov/jose2go v1.7.0 // indirect
 	github.com/emicklei/go-restful/v3 v3.11.0 // indirect
-	github.com/emirpasic/gods v1.12.0 // indirect
+	github.com/emirpasic/gods v1.18.1 // indirect
 	github.com/felixge/httpsnoop v1.0.4 // indirect
 	github.com/fsnotify/fsnotify v1.4.9 // indirect
 	github.com/fxamacker/cbor/v2 v2.7.0 // indirect
--- a/packages/gateway-v2/sql_handler.go
+++ b/packages/gateway-v2/sql_handler.go
@@ -58,9 +58,10 @@ func handleSQLRotateCredential(w http.ResponseWriter, r *http.Request) {
 	}
 	timeout := sqlDefaultTimeout
 	if env.TimeoutMs > 0 {
-		timeout = time.Duration(env.TimeoutMs) * time.Millisecond
-		if timeout > sqlMaxTimeout {
+		if int64(env.TimeoutMs) > int64(sqlMaxTimeout/time.Millisecond) {
 			timeout = sqlMaxTimeout
+		} else {
+			timeout = time.Duration(env.TimeoutMs) * time.Millisecond
 		}
 	}
 	ctx, cancel := context.WithTimeout(r.Context(), timeout)
@@ -123,7 +124,7 @@ func alterPasswordStatement(params sqlRotateParams) (string, error) {
 
 	stmt := fmt.Sprintf(`ALTER USER "%s" IDENTIFIED BY "%s"`, params.TargetUsername, params.NewPassword)
 
-	if strings.EqualFold(params.TargetUsername, params.Username) && params.Password != "" {
+	if params.TargetUsername == params.Username && params.Password != "" {
 		if strings.Contains(params.Password, `"`) {
 			return "", fmt.Errorf("oracle password cannot contain a double quote")
 		}
--- a/packages/gateway-v2/sql_handler.go
+++ b/packages/gateway-v2/sql_handler.go
@@ -1,0 +1,133 @@
+package gatewayv2
+
+import (
+	"bufio"
+	"context"
+	"crypto/tls"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"sync"
+	"time"
+)
+
+const (
+	maxSQLRequestBytes = 64 * 1024
+	sqlDefaultTimeout  = 30 * time.Second
+	sqlMaxTimeout      = 120 * time.Second
+	sqlReqDeadline     = 180 * time.Second
+)
+
+func serveSQLOverTLS(ctx context.Context, conn *tls.Conn, reader *bufio.Reader, forwardConfig *ForwardConfig) error {
+	return serveRPCOverTLS(ctx, conn, reader, forwardConfig, sqlMux(), sqlReqDeadline, "sql")
+}
+
+var sqlMux = sync.OnceValue(func() *http.ServeMux {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/v1/rotate-credential", handleSQLRotateCredential)
+	return mux
+})
+
+type sqlRotateParams struct {
+	sqlTestParams
+	TargetUsername string `json:"targetUsername"`
+	NewPassword    string `json:"newPassword"`
+}
+
+type sqlRotateEnvelope struct {
+	TimeoutMs int `json:"timeoutMs"`
+}
+
+func handleSQLRotateCredential(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		writeRPCError(w, http.StatusMethodNotAllowed, "Only POST is supported")
+		return
+	}
+	body, err := io.ReadAll(io.LimitReader(r.Body, maxSQLRequestBytes))
+	if err != nil {
+		writeRPCError(w, http.StatusBadRequest, "failed to read request body")
+		return
+	}
+
+	var env sqlRotateEnvelope
+	if err := json.Unmarshal(body, &env); err != nil {
+		writeRPCError(w, http.StatusBadRequest, "Invalid request body")
+		return
+	}
+	timeout := sqlDefaultTimeout
+	if env.TimeoutMs > 0 {
+		timeout = time.Duration(env.TimeoutMs) * time.Millisecond
+		if timeout > sqlMaxTimeout {
+			timeout = sqlMaxTimeout
+		}
+	}
+	ctx, cancel := context.WithTimeout(r.Context(), timeout)
+	defer cancel()
+
+	var params sqlRotateParams
+	if err := json.Unmarshal(body, &params); err != nil {
+		writeRPCError(w, http.StatusBadRequest, "Invalid request body")
+		return
+	}
+	if params.TargetUsername == "" || params.NewPassword == "" {
+		writeRPCError(w, http.StatusBadRequest, "targetUsername and newPassword are required")
+		return
+	}
+
+	statement, err := alterPasswordStatement(params)
+	if err != nil {
+		writeRPCError(w, http.StatusBadRequest, err.Error())
+		return
+	}
+
+	target, _ := r.Context().Value(rpcTargetContextKey{}).(rpcTarget)
+
+	if err := runWithContext(ctx, func() error {
+		return doSQLRotate(ctx, target.host, target.port, params, statement)
+	}); err != nil {
+		msg := redactSQLSecrets(err.Error(), params.Password, params.NewPassword)
+		writeRPCErrorWithKind(w, http.StatusBadGateway, msg, string(classifyTestConnFailure(err)))
+		return
+	}
+	writeRPCJSON(w, http.StatusOK, testConnectionResponse{Result: testConnectionResult{Ok: true}})
+}
+
+func doSQLRotate(ctx context.Context, host string, port int, params sqlRotateParams, statement string) error {
+	if err := dialTarget(ctx, host, port); err != nil {
+		return connectFailure(err)
+	}
+
+	db, err := openSQLTestDB(host, port, params.sqlTestParams)
+	if err != nil {
+		return connectFailure(err)
+	}
+	defer db.Close()
+	if _, err := db.ExecContext(ctx, statement); err != nil {
+		return authFailure(err)
+	}
+	return nil
+}
+
+func alterPasswordStatement(params sqlRotateParams) (string, error) {
+	if params.Dialect != "oracle" {
+		return "", fmt.Errorf("credential rotation over this transport is not supported for dialect %q", params.Dialect)
+	}
+
+	for label, v := range map[string]string{"username": params.TargetUsername, "password": params.NewPassword} {
+		if strings.Contains(v, `"`) {
+			return "", fmt.Errorf("oracle %s cannot contain a double quote", label)
+		}
+	}
+
+	stmt := fmt.Sprintf(`ALTER USER "%s" IDENTIFIED BY "%s"`, params.TargetUsername, params.NewPassword)
+
+	if strings.EqualFold(params.TargetUsername, params.Username) && params.Password != "" {
+		if strings.Contains(params.Password, `"`) {
+			return "", fmt.Errorf("oracle password cannot contain a double quote")
+		}
+		stmt += fmt.Sprintf(` REPLACE "%s"`, params.Password)
+	}
+	return stmt, nil
+}
--- a/packages/gateway-v2/sql_handler.go
+++ b/packages/gateway-v2/sql_handler.go
@@ -85,10 +85,10 @@ func handleSQLRotateCredential(w http.ResponseWriter, r *http.Request) {
 
 	target, _ := r.Context().Value(rpcTargetContextKey{}).(rpcTarget)
 
-	if err := runWithContext(ctx, func() error {
+	if err := runWithContextTimeoutMessage(ctx, func() error {
 		return doSQLRotate(ctx, target.host, target.port, params, statement)
-	}); err != nil {
-		msg := redactSQLSecrets(err.Error(), params.Password, params.NewPassword)
+	}, "the password change timed out before the target answered, and may still have been applied"); err != nil {
+		msg := redactProbeSecrets(err.Error(), params.Password, params.NewPassword)
 		writeRPCErrorWithKind(w, http.StatusBadGateway, msg, string(classifyTestConnFailure(err)))
 		return
 	}
@@ -106,7 +106,7 @@ func doSQLRotate(ctx context.Context, host string, port int, params sqlRotatePar
 	}
 	defer db.Close()
 	if _, err := db.ExecContext(ctx, statement); err != nil {
-		return authFailure(err)
+		return sqlAuthFailure(params.Dialect, err)
 	}
 	return nil
 }
--- a/packages/gateway-v2/test_connection_handler.go
+++ b/packages/gateway-v2/test_connection_handler.go
@@ -216,7 +216,7 @@ func openSQLTestDB(host string, port int, params sqlTestParams) (*sql.DB, error)
 		}
 		connStr := go_ora.BuildUrl(host, port, params.Database, params.Username, params.Password, urlOptions)
 		if _, err := configurations.ParseConfig(connStr); err != nil {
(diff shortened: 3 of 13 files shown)


packages/gateway-v2/sql_handler.go
```diff
@@ -0,0 +1,134 @@
...
+	if err := dialTarget(ctx, host, port); err != nil {
+		return connectFailure(err)
+	}
+
+	db, err := openSQLTestDB(host, port, params.sqlTestParams)
+	if err != nil {
+		return connectFailure(err)
+	}
+	defer db.Close()
+	if _, err := db.ExecContext(ctx, statement); err != nil {
```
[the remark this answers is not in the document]
carlosmonastyrski: Correct that the statement carries the password, and the mitigation is to enable SSL on the account, which Oracle PAM supports

carlosmonastyrski:
@greptile re-review this PR and update the top Greptile Summary

packages/gateway-v2/test_connection_handler.go
```diff
@@ -634,8 +665,22 @@ func handleTestConnection(w http.ResponseWriter, r *http.Request) {
...
 
 	if testErr := runWithContext(ctx, op); testErr != nil {
-		writeRPCErrorWithKind(w, http.StatusBadGateway, testErr.Error(), string(classifyTestConnFailure(testErr)))
+		writeRPCErrorWithKind(w, http.StatusBadGateway, redactSQLSecrets(testErr.Error(), redactSecrets...), string(classifyTestConnFailure(testErr)))
 		return
 	}
 	writeRPCJSON(w, http.StatusOK, testConnectionResponse{Result: testConnectionResult{Ok: true}})
 }
+
+var urlUserinfoPattern = regexp.MustCompile(`(?i)([a-z][a-z0-9+.\-]*://)[^/\s@]*@`)
```
lb-vn: `[^/\s@]*` cannot cross an @, so it stops at the first one and a password containing @ only gets half redacted: `oracle://system:p@ssw0rd@host:1521/svc` comes out as `oracle://******@ssw0rd@host:1521/svc`. it does not bite for sql because params.Password is passed explicitly, but mongo/redis/ldap/ssh pass no secrets at all and this regex is the only thing covering them. making it greedy, `[^/\s]*@`, matches up to the last @ before the host and still cannot cross a / or whitespace, so it is strictly better here.

packages/gateway-v2/test_connection_handler.go
```diff
@@ -578,12 +607,14 @@ func handleTestConnection(w http.ResponseWriter, r *http.Request) {
...
 
 	var op func() error
+	var redactSecrets []string
 	switch env.Mode {
 	case testConnModeSQL:
 		var params sqlTestParams
 		if !decode(&params) {
 			return
 		}
+		redactSecrets = append(redactSecrets, params.Password)
```
lb-vn: only the sql mode populates redactSecrets, the other five still hand testErr.Error() back with nothing but the url regex covering them. worth doing the same for mongo/redis/ldap/ssh while we are in here, or a follow-up.

packages/gateway-v2/test_connection_handler.go
```diff
@@ -204,11 +209,35 @@ func openSQLTestDB(host string, port int, params sqlTestParams) (*sql.DB, error)
 			config.TLSConfig = tlsConfig
 		}
 		return sql.OpenDB(mssql.NewConnectorConfig(config)), nil
+	case "oracle":
+		urlOptions := map[string]string{}
+		if params.SslEnabled {
+			urlOptions["SSL"] = "TRUE"
+		}
+		connStr := go_ora.BuildUrl(host, port, params.Database, params.Username, params.Password, urlOptions)
+		if _, err := configurations.ParseConfig(connStr); err != nil {
```
lb-vn: took me a minute to work out why we parse the connection string and throw the result away. as far as i can tell NewConnector swallows the parse error into a lazily failing connector, so this is the only point where we can surface it with the password redacted. one line saying that would save the next reader.

packages/gateway-v2/sql_handler.go
```diff
@@ -0,0 +1,134 @@
...
+
+	statement, err := alterPasswordStatement(params)
+	if err != nil {
+		writeRPCError(w, http.StatusBadRequest, err.Error())
+		return
+	}
+
+	target, _ := r.Context().Value(rpcTargetContextKey{}).(rpcTarget)
+
+	if err := runWithContext(ctx, func() error {
```
lb-vn: on timeout runWithContext abandons the goroutine and returns "connection test timed out", so for a rotation the ALTER USER can still land after we have already reported failure, and the message the user ends up reading talks about a connection test. the platform side covers the first half via the staged pending credential, but the wording still surfaces in the rotation error. we should give the rotate path its own message.

packages/pam/local/access.go
```diff
@@ -304,8 +305,9 @@ func handleApprovalRequired(httpClient *resty.Client, err error, path, reason, d
 // instructions handed to an agent are built from these, so the two cannot describe the same account
 // differently, and a fix to how one connects is a fix for both.
 type AccountConnectionDisplay struct {
-	TypeLabel   string // e.g., "PostgreSQL", "MySQL", "SQL Server"
-	DefaultPort int    // default port for this account type
+	TypeLabel        string // e.g., "PostgreSQL", "MySQL", "SQL Server"
+	DefaultPort      int    // default port for this account type
+	RequiredPassword string
```
lb-vn: every other field in this struct is commented and this is the one that most needs it, since it is not the account's password, it is a literal placeholder the proxy swaps out. one line.

go.mod
```diff
@@ -104,7 +105,7 @@ require (
 	github.com/dustin/go-humanize v1.0.0 // indirect
 	github.com/dvsekhvalnov/jose2go v1.7.0 // indirect
 	github.com/emicklei/go-restful/v3 v3.11.0 // indirect
-	github.com/emirpasic/gods v1.12.0 // indirect
+	github.com/emirpasic/gods v1.18.1 // indirect
```
lb-vn: emirpasic/gods jumping 1.12.0 to 1.18.1 looks like tidy fallout rather than something go-ora needs. intentional?

packages/gateway-v2/test_connection_handler.go
lb-vn: `[^/\s@]*` cannot cross an @, so it stops at the first one and a password containing @ only gets half redacted: `oracle://system:p@ssw0rd@host:1521/svc` comes out as `oracle://******@ssw0rd@host:1521/svc`. it does not bite for sql because params.Password is passed explicitly, but mongo/redis/ldap/ssh pass no secrets at all and this regex is the only thing covering them. making it greedy, `[^/\s]*@`, matches up to the last @ before the host and still cannot cross a / or whitespace, so it is strictly better here.

packages/gateway-v2/test_connection_handler.go
lb-vn: only the sql mode populates redactSecrets, the other five still hand testErr.Error() back with nothing but the url regex covering them. worth doing the same for mongo/redis/ldap/ssh while we are in here, or a follow-up.

packages/gateway-v2/test_connection_handler.go
lb-vn: took me a minute to work out why we parse the connection string and throw the result away. as far as i can tell NewConnector swallows the parse error into a lazily failing connector, so this is the only point where we can surface it with the password redacted. one line saying that would save the next reader.

packages/gateway-v2/sql_handler.go
lb-vn: on timeout runWithContext abandons the goroutine and returns "connection test timed out", so for a rotation the ALTER USER can still land after we have already reported failure, and the message the user ends up reading talks about a connection test. the platform side covers the first half via the staged pending credential, but the wording still surfaces in the rotation error. we should give the rotate path its own message.

packages/pam/local/access.go
lb-vn: every other field in this struct is commented and this is the one that most needs it, since it is not the account's password, it is a literal placeholder the proxy swaps out. one line.

go.mod
lb-vn: emirpasic/gods jumping 1.12.0 to 1.18.1 looks like tidy fallout rather than something go-ora needs. intentional?

packages/gateway-v2/test_connection_handler.go
```diff
@@ -246,7 +275,7 @@ func doSQLConnectionTest(ctx context.Context, host string, port int, params sqlT
 	defer db.Close()
 
 	var result int
-	return authFailure(db.QueryRowContext(ctx, "SELECT 1").Scan(&result))
+	return authFailure(db.QueryRowContext(ctx, sqlVerifyQuery(params.Dialect)).Scan(&result))
```
lb-vn: tested this end to end against oracle 23.8 through a real gateway and it misclassifies a whole class of oracle failure as a rejected credential.

go-ora connects lazily, so the TNS service resolution happens inside this QueryRowContext, not in dialTarget above. an oracle service that the listener does not know returns `ORA-12514: TNS:listener does not currently know of service requested in connect descriptor` from right here, authFailure sees it is not a net.Error, and tags it `auth`. i created a pam account with a typo'd service name and the health check came back `invalid-credentials` / Out of Sync quoting ORA-12514 verbatim.

that is worse than a cosmetic mislabel, because statusForFailureKind maps auth to InvalidCredentials and stopsSchedule halts the heartbeat on it by design. so a closed or restarting PDB, a failed-over service, or a service-name typo permanently stops checks on the account and tells the operator their password is wrong when it is not.

the phase-tagging comment in test_connection_failure_kind.go says each probe tags the phase it failed in rather than reading the classification out of the driver's error text, and for the other dialects that holds because connect and auth are separable. for oracle they are not: one call does service resolution, handshake and auth. so this is the one dialect where we do have to look at the code. suggest mapping the listener/service-side ORA codes (12514, 12541, 12537, and 01033 for a database still mounting) to failureKindTransport before falling through to authFailure. postgres has the same shape with a bad dbname, but there the database is a secondary field and a typo is a one-time setup error, whereas for oracle the service name is the connection identity and these codes are routine and transient.

lb-vn: Claude:

tested against real oracle: a wrong service name returns ORA-12514 from this call, and authFailure tags it as auth. that marks the account Out of Sync and stops the heartbeat, so a closed PDB or a typo permanently stops checks and blames the password. oracle does service resolution and auth in one call so the phase split doesn't work here. we should map 12514, 12541, 12537 and 01033 to transport before falling through.

packages/gateway-v2/test_connection_handler.go
lb-vn: Claude:

this stops at the first @, so a password containing @ only gets half redacted. sql is fine because we pass the password explicitly, but the other modes rely on this alone. `[^/\s]*@` fixes it.

packages/gateway-v2/test_connection_handler.go
lb-vn, line 621: only sql passes secrets here, mongo/redis/ldap/ssh pass nothing. we should do the same for those.

packages/gateway-v2/test_connection_handler.go
lb-vn: we should add a line saying why we parse and throw it away: NewConnector hides the parse error, so this is the only place to catch it with the password redacted.

packages/gateway-v2/sql_handler.go
lb-vn: on timeout this says "connection test timed out" and the ALTER can still land afterwards. we should give rotation its own message.

packages/pam/local/access.go
lb-vn, line 313: we should add a comment explaining what RequiredPassword is

go.mod
lb-vn, line 108: Claude:

is this gods bump intentional? looks unrelated to go-ora.

lb-vn (changes requested):

go.mod
carlosmonastyrski: yes, go-ora/v2 v2.9.0 requires gods v1.18.1

packages/pam/handlers/oracle/proxy_auth.go
```diff
@@ -151,58 +151,58 @@ func (p *OracleProxy) handleConnectionProxied(ctx context.Context, clientConn ne
...
 	log.Info().Str("sessionID", p.config.SessionID).Int("p1Len", len(p1Payload)).Msg("Proxy: auth-request boundary reached")
 
 	p1Forward := p1Payload
 	if p.config.InjectUsername != "" {
-		rewritten, rerr := rewritePhase1User(p1Payload, p.config.InjectUsername)
-		if rerr != nil {
-			return fmt.Errorf("rewrite phase 1 username: %w", rerr)
+		rewritten, applied := rewriteBundledUsername(p1Payload, AuthSubOpPhaseOne, p.config.InjectUsername)
+		if !applied {
+			log.Warn().Str("sessionID", p.config.SessionID).Msg("Proxy: this client encodes the auth request in a layout the proxy cannot rewrite; the username it sent is forwarded as is and Oracle rejects it unless it names the account")
```
lb-vn: the sqlplus hang is fixed, i verified it authenticates through the proxy now. but this warning fires on every sqlplus connection, on both phases. i instrumented rewriteAuthRequestUser and it returns "auth request carries no username" at offset 0 for both 0x76 and 0x73, so InjectUsername is inert for oci clients.

it only worked for me because i typed APP_USER, which is what the banner prints. anything else gets a bare ORA-01017 with the real reason only in the gateway log. it does fail closed so it isn't a security problem, but the old code aborted on a failed rewrite and this one carries on. we should fail with something the client can see rather than depending on the user typing the right name.

packages/pam/handlers/oracle/proxy_auth.go
```diff
@@ -151,58 +151,58 @@ func (p *OracleProxy) handleConnectionProxied(ctx context.Context, clientConn ne
...
 		}
 		p2ReqTranslated = rewritten
 	}
-	if err := writeDataPayload(upstreamConn, p2ReqTranslated, use32Bit); err != nil {
+	if err := writeDataPacket(upstreamConn, &DataPacket{DataFlag: p2ReqPkt.DataFlag, Payload: p2ReqTranslated}, use32Bit); err != nil {
 		return fmt.Errorf("forward phase 2 request: %w", err)
 	}
 	log.Info().Str("sessionID", p.config.SessionID).Msg("Proxy: phase-2 request translated and forwarded")
```
lb-vn: IGNORE PLACEMENT

the "phase-2 response forwarded; client authenticated" log below claims success even when oracle rejected. i connected as SYSTEM with the placeholder password, got ORA-01017, and it still logged authenticated. we should look at the response before saying that.

packages/gateway-v2/sql_handler.go
```diff
@@ -0,0 +1,201 @@
...
+		}
+		stmt += fmt.Sprintf(` REPLACE "%s"`, params.Password)
+	}
+	return stmt, nil
+}
+
+func resolveOracleUsername(ctx context.Context, db *sql.DB, name string) (resolved string, exact bool, err error) {
+	rows, qerr := db.QueryContext(ctx, `SELECT username FROM all_users WHERE UPPER(username) = UPPER(:1)`, name)
+	if qerr != nil {
+		return name, false, nil
```
lb-vn: oracle uppercases a username unless it was created with quotes, so someone who typed `app_user` into infisical is really `APP_USER` in the database. that mismatch is what this lookup exists to fix.

if the lookup fails we keep the name as typed and the rotation runs `ALTER USER "app_user"`. the quotes make oracle look for a literally lowercase user, it doesn't find one, and rotation fails with ORA-01918 user does not exist.

all four exits here do that without a word, so a permissions or connection problem on the lookup is indistinguishable from the user genuinely not existing. we should log when the lookup fails.

packages/gateway-v2/sql_handler.go
```diff
@@ -0,0 +1,201 @@
...
+		}
+		matches = append(matches, found)
+	}
+	if rerr := rows.Err(); rerr != nil {
+		return name, false, nil
+	}
+
+	switch len(matches) {
+	case 0:
+		return name, false, nil
```
lb-vn: we know the user doesn't exist at this point, we just looked. falling through so oracle can answer ORA-01918 works, but we can say it better here.

packages/pam/handlers/oracle/oci_auth.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	first := -1
+	for offset := 0; offset+1 < firstKey; offset++ {
+		if payload[offset] != TTCMsgAuthRequest || payload[offset+1] != expectedSubOp {
+			continue
+		}
+		if first < 0 {
+			first = offset
+		}
+		if _, err := rewriteAuthRequestUser(payload[offset:], expectedSubOp, "PROBE"); err == nil {
```
lb-vn: this builds a full rewritten payload at every candidate offset just to test the offset, and ociContainsAuthRequest calls it for every client packet in the pre-auth loop. we should validate without allocating the rewrite.

packages/pam/handlers/oracle/proxy_auth.go
lb-vn: this warning fires on every sqlplus connection, on both phases, so InjectUsername never actually applies for sqlplus.

it still connects only because the user types the username the banner prints. anything else gets a bare ORA-01017. the old code aborted when the rewrite failed, this one carries on. we should fail with an error the client can see.

packages/pam/handlers/oracle/proxy_auth.go
lb-vn: IGNORE PLACEMENT

the "phase-2 response forwarded; client authenticated" log below claims success even when oracle rejected. i connected as SYSTEM with the placeholder password, got ORA-01017, and it still logged authenticated. we should look at the response before saying that.

packages/gateway-v2/sql_handler.go
lb-vn: I think we should log in every error block before returning the provided username just as a way to add visibility instead of silently defaulting back. This would apply to the blocks below as well

packages/gateway-v2/sql_handler.go
lb-vn: here we know that the user definitely does not exist, so I wonder if returning the name is the right call? it just defers an error to elsewhere

packages/pam/handlers/oracle/proxy_auth.go
lb-vn: this fires on every sqlplus connection, so InjectUsername never actually applies for sqlplus. rewriteAuthRequestUser returns "auth request carries no username", so the oci layout isn't being parsed.

we should make the rewrite handle that layout. if it can't be, this should fail the connection rather than carry on with whatever the user typed.

packages/pam/handlers/oracle/proxy_auth.go
lb-vn, line 210: IGNORE PLACEMENT

Claude:

the "phase-2 response forwarded; client authenticated" log below claims success even when oracle rejected. i connected as SYSTEM with the placeholder password, got ORA-01017, and it still logged authenticated. we should look at the response before saying that.

packages/gateway-v2/sql_handler.go
lb-vn, line 169: I think we should log in every error block before returning the provided username just as a way to add visibility instead of silently defaulting back. This would apply to the blocks below as well

packages/gateway-v2/sql_handler.go
lb-vn: here we know that the user definitely does not exist, so I wonder if returning the name is the right call? it just defers an error to elsewhere

lb-vn (changes requested):

lb-vn (approved):

PR #392, opened 2026-09-08
fix(pam): stop retrying rejected recording chunks

bernie-g:
# Description 📣

Chunk reconciliation retried every failed upload on a 5-minute tick with no give-up, so a chunk the platform rejects for a reason that cannot change was re-POSTed for the life of the gateway. One stuck queue produced 22,511 errors in three hours in production, and the burst also saturated the write rate limit.

Fixes the gateway half of [PAM-463](https://linear.app/infisical/issue/PAM-463/investigate-session-recording-endpoint-error-spike); the platform half is in Infisical/infisical#8043. This is the only change that clears an already-stuck queue, so it needs to reach affected gateways for the error rate to drop.

## Type ✨

- [x] Bug fix

# Tests 🛠️

Four new tests drive `UploadChunk` against a real `httptest` server with `config.INFISICAL_URL` overridden, so the status code travels the genuine production path (resty response → `NewAPIErrorWithResponse` → `fmt.Errorf("%w")` → `errors.As`) rather than a synthetic error:

- a real `400 Invalid upload token` drops the chunk
- a real `429` and a real `500` both keep it queued
- a queue of five permanently rejected chunks drains in one reconcile pass and makes **zero** requests on the second pass

The second commit adds `singleflight` around session credential fetches. Recording secrets are minted on a session's first fetch, and the cache check released its lock before the API call, so two connections on one session both fetched and both minted, leaving this gateway holding a token the platform had discarded. Covered by a test that fires 16 concurrent callers and asserts one fetch, run under `-race`.

```sh
go test ./packages/pam/... 
go test ./packages/pam/session/ -run "Real|DrainsPermanently|Cached|Singleflight" -race -v
```

Also built the binary from this branch and ran it as a real gateway against a local platform: it enrolls, connects to the relay, heartbeats, and starts the session uploader with no errors.

---

### DIFF
--- a/packages/pam/session/chunk_uploader.go
+++ b/packages/pam/session/chunk_uploader.go
@@ -9,6 +9,7 @@ import (
 	"crypto/sha256"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -263,6 +264,41 @@ func (cu *ChunkUploader) EncryptAndQueueChunk(
 	return pc, nil
 }
 
+// Reconciliation never gives up, so a chunk that can only ever be rejected retries forever.
+func isPermanentUploadFailure(err error) bool {
+	var apiErr *api.APIError
+	if !errors.As(err, &apiErr) {
+		return false
+	}
+	switch apiErr.StatusCode {
+	case http.StatusBadRequest, http.StatusForbidden, http.StatusNotFound:
+		return true
+	default:
+		return false
+	}
+}
+
+// flushSession has already advanced past it, so keeping the file only feeds the retry loop.
+func (cu *ChunkUploader) dropIfPermanent(sessionID string, pc *pendingChunk, err error) {
+	if !isPermanentUploadFailure(err) {
+		return
+	}
+
+	entry := log.Error().
+		Err(err).
+		Str("sessionId", sessionID).
+		Int("chunkIndex", pc.ChunkIndex).
+		Int("ciphertextBytes", len(pc.Ciphertext))
+
+	if rmErr := os.Remove(chunkPendingFile(sessionID, pc.ChunkIndex)); rmErr != nil {
+		entry.AnErr("removeError", rmErr).
+			Msg("Recording chunk permanently rejected but could not be removed; it stays queued and will keep failing")
+		return
+	}
+
+	entry.Msg("Recording chunk permanently rejected by platform; dropped instead of retrying")
+}
+
 func (cu *ChunkUploader) UploadChunk(sessionID string, pc *pendingChunk) error {
 	secrets := cu.credentialsManager.GetRecordingSecrets(sessionID)
 	if secrets == nil {
@@ -281,6 +317,7 @@ func (cu *ChunkUploader) UploadChunk(sessionID string, pc *pendingChunk) error {
 			},
 		)
 		if err != nil {
+			cu.dropIfPermanent(sessionID, pc, err)
 			return fmt.Errorf("presigned PUT mint failed: %w", err)
 		}
 		if err := s3PutCiphertext(presigned.URL, pc.Ciphertext); err != nil {
@@ -305,6 +342,7 @@ func (cu *ChunkUploader) UploadChunk(sessionID string, pc *pendingChunk) error {
 	}
 
 	if err := api.CallPAMSessionChunkMetadata(cu.httpClient, sessionID, secrets.UploadToken, metadataReq); err != nil {
+		cu.dropIfPermanent(sessionID, pc, err)
 		return fmt.Errorf("chunk metadata POST failed: %w", err)
 	}
 
--- a/packages/pam/session/chunk_uploader.go
+++ b/packages/pam/session/chunk_uploader.go
@@ -264,18 +264,16 @@ func (cu *ChunkUploader) EncryptAndQueueChunk(
 	return pc, nil
 }
 
-// Reconciliation never gives up, so a chunk that can only ever be rejected retries forever.
+const uploadTokenMismatchError = "PamUploadTokenHashMismatch"
+
+// Matched on the platform's error name rather than the status, because a 404 or a
+// missing-token 400 can just be a read replica that has not caught up with a new session.
 func isPermanentUploadFailure(err error) bool {
 	var apiErr *api.APIError
 	if !errors.As(err, &apiErr) {
 		return false
 	}
-	switch apiErr.StatusCode {
-	case http.StatusBadRequest, http.StatusForbidden, http.StatusNotFound:
-		return true
-	default:
-		return false
-	}
+	return apiErr.StatusCode == http.StatusBadRequest && apiErr.Name == uploadTokenMismatchError
 }
 
 // flushSession has already advanced past it, so keeping the file only feeds the retry loop.
--- a/packages/pam/session/chunk_uploader_http_test.go
+++ b/packages/pam/session/chunk_uploader_http_test.go
@@ -1,0 +1,140 @@
+package session
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
+
+	"github.com/Infisical/infisical-merge/packages/config"
+	"github.com/go-resty/resty/v2"
+)
+
+// Uses a real HTTP server so the status code travels the same path it does in production.
+func newChunkUploaderAgainst(t *testing.T, status int, body string) (*ChunkUploader, *int) {
+	t.Helper()
+
+	hits := 0
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		hits += 1
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(status)
+		_, _ = w.Write([]byte(body))
+	}))
+	t.Cleanup(srv.Close)
+
+	orig := config.INFISICAL_URL
+	config.INFISICAL_URL = srv.URL + "/api"
+	t.Cleanup(func() { config.INFISICAL_URL = orig })
+
+	cm := newTestCredentialsManager(t)
+	cm.recordingSecrets["sess"] = &PAMRecordingSecrets{
+		SessionKey:     make([]byte, 32),
+		UploadToken:    "fake-token-for-test",
+		StorageBackend: storageBackendPostgres,
+		ProjectId:      "proj",
+		SessionId:      "sess",
+	}
+
+	return NewChunkUploader(resty.New(), cm), &hits
+}
+
+func queuedChunk(t *testing.T, sessionID string) *pendingChunk {
+	t.Helper()
+	pc := &pendingChunk{
+		ChunkIndex:     0,
+		StartElapsedMs: 0,
+		EndElapsedMs:   10,
+		IV:             make([]byte, 12),
+		Ciphertext:     []byte("ciphertext"),
+		Sha256:         make([]byte, 32),
+		StorageBackend: storageBackendPostgres,
+	}
+	if err := writePendingChunk(sessionID, pc); err != nil {
+		t.Fatal(err)
+	}
+	return pc
+}
+
+func TestUploadChunk_RealInvalidUploadTokenIsDropped(t *testing.T) {
+	setupTestDir(t)
+
+	cu, hits := newChunkUploaderAgainst(t, http.StatusBadRequest,
+		`{"reqId":"req-x","statusCode":400,"message":"Invalid upload token","error":"PamUploadTokenHashMismatch"}`)
+	pc := queuedChunk(t, "sess")
+
+	if err := cu.UploadChunk("sess", pc); err == nil {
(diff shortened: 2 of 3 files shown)


packages/pam/session/chunk_uploader.go
```diff
@@ -263,6 +264,42 @@ func (cu *ChunkUploader) EncryptAndQueueChunk(
...
+	if !errors.As(err, &apiErr) {
+		return false
+	}
+	switch apiErr.StatusCode {
+	case http.StatusBadRequest,
+		http.StatusForbidden,
+		http.StatusNotFound,
+		http.StatusConflict,
+		http.StatusRequestEntityTooLarge,
+		http.StatusUnprocessableEntity:
```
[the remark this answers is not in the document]
bernie-g: Asking the server for the token again never returns a new one, so a retry would fail exactly the same way.

packages/pam/session/chunk_uploader.go
```diff
@@ -263,6 +264,42 @@ func (cu *ChunkUploader) EncryptAndQueueChunk(
...
+	if !isPermanentUploadFailure(err) {
+		return
+	}
+	log.Error().
+		Err(err).
+		Str("sessionId", sessionID).
+		Int("chunkIndex", pc.ChunkIndex).
+		Int("ciphertextBytes", len(pc.Ciphertext)).
+		Msg("Recording chunk permanently rejected by platform; dropping it instead of retrying")
+	deletePendingChunk(sessionID, pc.ChunkIndex)
```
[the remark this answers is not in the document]
bernie-g: Fixed

packages/pam/session/credentials_test.go
```diff
@@ -0,0 +1,72 @@
...
+	var started, done sync.WaitGroup
+	started.Add(goroutines)
+	done.Add(goroutines)
+
+	for i := 0; i < goroutines; i += 1 {
+		go func() {
+			defer done.Done()
+			started.Done()
+			// Mirrors GetPAMSessionCredentials: every caller goes through the same singleflight key.
+			_, _, _ = cm.credentialsFetch.Do("session-a", func() (any, error) {
```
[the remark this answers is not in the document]
bernie-g: That change is no longer in this PR, so there is nothing left to test.

packages/pam/session/chunk_uploader.go
```diff
@@ -263,6 +264,48 @@ func (cu *ChunkUploader) EncryptAndQueueChunk(
...
+// These statuses describe the chunk's own content or its session's state, so the same
+// chunk is rejected on every future attempt. Reconciliation has no give-up, so without
+// this one rejected chunk becomes a permanent stream of API errors.
+func isPermanentUploadFailure(err error) bool {
+	var apiErr *api.APIError
+	if !errors.As(err, &apiErr) {
+		return false
+	}
+	switch apiErr.StatusCode {
+	case http.StatusBadRequest, http.StatusForbidden, http.StatusNotFound:
```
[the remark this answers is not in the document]
bernie-g: This is how it already behaves today and is not caused by this change, so it is tracked separately.

packages/pam/session/chunk_uploader.go
```diff
@@ -263,6 +264,42 @@ func (cu *ChunkUploader) EncryptAndQueueChunk(
...
+	var apiErr *api.APIError
+	if !errors.As(err, &apiErr) {
+		return false
+	}
+	switch apiErr.StatusCode {
+	case http.StatusBadRequest,
+		http.StatusForbidden,
+		http.StatusNotFound,
+		http.StatusConflict,
+		http.StatusRequestEntityTooLarge,
```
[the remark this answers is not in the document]
bernie-g: We no longer delete on a size error, so nothing is thrown away.

packages/pam/session/chunk_uploader.go
```diff
@@ -74,6 +75,21 @@ func chunkPendingFile(sessionID string, chunkIndex int) string {
...
+	return filepath.Join(GetSessionRecordingDir(), "rejected-chunks", sessionID)
+}
+
+func quarantinePendingChunk(sessionID string, chunkIndex int) error {
+	dir := chunkRejectedDir(sessionID)
+	if err := os.MkdirAll(dir, 0o755); err != nil {
+		return fmt.Errorf("create rejected chunk dir: %w", err)
+	}
+	dst := filepath.Join(dir, fmt.Sprintf("%06d.chunk", chunkIndex))
+	return os.Rename(chunkPendingFile(sessionID, chunkIndex), dst)
```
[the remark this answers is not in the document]
bernie-g: The quarantine was removed, so nothing builds up any more.

packages/pam/session/chunk_uploader.go
```diff
@@ -263,6 +264,41 @@ func (cu *ChunkUploader) EncryptAndQueueChunk(
...
 }
 
+// Reconciliation never gives up, so a chunk that can only ever be rejected retries forever.
+func isPermanentUploadFailure(err error) bool {
+	var apiErr *api.APIError
+	if !errors.As(err, &apiErr) {
+		return false
+	}
+	switch apiErr.StatusCode {
+	case http.StatusBadRequest, http.StatusForbidden, http.StatusNotFound:
```
lb-vn: we should not treat 404 as permanent. on the platform side `recordChunk` calls `pamSessionDAL.findById`, which reads from `db.replicaNode()`, and it throws `Session <id> not found` before it checks anything else. so a session row that hasn't replicated yet gives us a 404 and we now delete the chunk instead of retrying it - that's the opening chunks of a recording gone, silently, on an audit feature. worth noting the DAL comment in Infisical/infisical#8043 explicitly cites lagging replicas as the reason the claim reads from the primary, so we already know that read path can lie to us. same concern for a 404 from a proxy or a version-skewed self-hosted platform where the route isn't registered yet.

packages/pam/session/chunk_uploader.go
```diff
@@ -263,6 +264,41 @@ func (cu *ChunkUploader) EncryptAndQueueChunk(
 	return pc, nil
 }
 
+// Reconciliation never gives up, so a chunk that can only ever be rejected retries forever.
+func isPermanentUploadFailure(err error) bool {
+	var apiErr *api.APIError
+	if !errors.As(err, &apiErr) {
+		return false
+	}
+	switch apiErr.StatusCode {
```
lb-vn: 403 has a transient source too. the platform's own 403 is `Gateway not authorized for this session` on a gatewayId/identityId mismatch, which is fair to drop. but a WAF or cloudflare challenge sitting in front of the api also returns 403, and that would wipe every gateway's queue at once. since #8043 just gave us distinct error names, we should key the drop on the platform's error name rather than the bare status.

packages/pam/session/chunk_uploader.go
```diff
@@ -305,6 +342,7 @@ func (cu *ChunkUploader) UploadChunk(sessionID string, pc *pendingChunk) error {
 	}
 
 	if err := api.CallPAMSessionChunkMetadata(cu.httpClient, sessionID, secrets.UploadToken, metadataReq); err != nil {
+		cu.dropIfPermanent(sessionID, pc, err)
```
lb-vn: this one interacts badly with the platform half. the losing fetch in #8043 returns `uploadToken: ""`, and credentials.go only backfills an empty token from an entry that's *already* cached - so if the loser's response lands before the winner's there's nothing to fall back on and we cache an empty token. any chunk uploaded in that window gets `PamUploadTokenMissing` 400, and we now delete it. before this change that chunk just retried and succeeded on the next tick. between the two prs we'd be turning a millisecond transient into permanent data loss.

we should bound this rather than dropping on first rejection: either move the file to a `.rejected` dir so it's recoverable, or re-fetch credentials once and only give up on the second rejection. if we want to stay narrow, gating on `apiErr.Name == "PamUploadTokenHashMismatch"` gets the stuck-queue case without catching every other 400.

packages/pam/session/chunk_uploader.go
```diff
@@ -263,6 +264,41 @@ func (cu *ChunkUploader) EncryptAndQueueChunk(
...
+	}
+	switch apiErr.StatusCode {
+	case http.StatusBadRequest, http.StatusForbidden, http.StatusNotFound:
+		return true
+	default:
+		return false
+	}
+}
+
+// flushSession has already advanced past it, so keeping the file only feeds the retry loop.
```
lb-vn: this comment only holds for the reconcile path. `UploadChunk` is also called from `flushSession` at uploader.go:607, so a permanent rejection there drops a chunk out of a session that's still live and leaves a hole in the middle of an otherwise-valid recording. do we know playback tolerates a missing chunk index, and should we be marking the recording incomplete when we drop one?

packages/pam/session/chunk_uploader_test.go
```diff
@@ -346,6 +350,62 @@ func TestDeletePendingChunk(t *testing.T) {
 	}
 }
 
+func TestIsPermanentUploadFailure(t *testing.T) {
+	cases := []struct {
+		name string
+		err  error
+		want bool
+	}{
+		{"body too large is retryable after re-chunking", &api.APIError{StatusCode: http.StatusRequestEntityTooLarge}, false},
```
lb-vn: nothing re-chunks a chunk that's already written to the queue, so a 413 here is exactly the forever-loop this pr exists to fix, just with a different status. either the case name should say we're knowingly leaving that one, or 413 belongs in the permanent set.

packages/pam/session/chunk_uploader_http_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+	"github.com/Infisical/infisical-merge/packages/config"
+	"github.com/go-resty/resty/v2"
+)
+
+// Uses a real HTTP server so the status code travels the same path it does in production.
+func newChunkUploaderAgainst(t *testing.T, status int, body string) (*ChunkUploader, *int) {
+	t.Helper()
+
+	hits := 0
```
lb-vn: nit: `hits` is a plain int written from the httptest handler goroutine. fine while every test drives requests serially, but it'll trip `-race` the first time someone reuses this helper for a concurrent case. `atomic.Int64` costs nothing here.

packages/pam/session/chunk_uploader.go
lb-vn: Claude:

404 shouldn't be permanent. the platform looks the session up on a read replica, so a row that hasn't replicated yet 404s and we'd delete the chunk instead of retrying. that's the start of the recording gone.

packages/pam/session/chunk_uploader.go
lb-vn: 403 also comes from a WAF or cloudflare challenge, not just the platform's gateway mismatch check, and that would wipe every gateway's queue at once. can we key off the error name instead of the bare status?

packages/pam/session/chunk_uploader.go
lb-vn: the loser in the platform pr gets an empty upload token, and if its response lands before the winner's we cache that. any chunk uploaded in that window 400s and we'd now delete it instead of retrying. we should move these to a rejected dir, or re-fetch credentials once before giving up.

packages/pam/session/chunk_uploader.go
lb-vn: UploadChunk also runs from flushSession, so this drops chunks out of live sessions too and leaves a hole mid-recording. does playback handle a missing index?

packages/pam/session/chunk_uploader_test.go
lb-vn: nothing re-chunks a chunk that's already queued, so a 413 loops forever - same bug this pr is fixing.

packages/pam/session/chunk_uploader_http_test.go
lb-vn: hits is written from the handler goroutine, will trip -race if this helper ever gets used concurrently. atomic.Int64?

lb-vn (changes requested):

packages/pam/session/chunk_uploader.go
bernie-g: Fixed

packages/pam/session/chunk_uploader.go
```diff
@@ -263,6 +264,39 @@ func (cu *ChunkUploader) EncryptAndQueueChunk(
...
+		return
+	}
+
+	entry := log.Error().
+		Err(err).
+		Str("sessionId", sessionID).
+		Int("chunkIndex", pc.ChunkIndex).
+		Int("ciphertextBytes", len(pc.Ciphertext))
+
+	if rmErr := os.Remove(chunkPendingFile(sessionID, pc.ChunkIndex)); rmErr != nil {
```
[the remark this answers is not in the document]
bernie-g: Keeping the chunk locally would not preserve the recording either, so we are leaving this out to keep the change small.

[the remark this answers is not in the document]
bernie-g: This doesn't seem like a big issue.

lb-vn (approved):

PR #397, opened 2026-09-15
feat(agent-vault): enforce method and path rules, inject custom headers, substitute placeholders

saifsmailbox98:
# Description 📣

Proxy side of method and path rules, custom headers and substitutions. The API hands the resolved service its `allowedMethods`, `allowedPathPrefixes`, `customHeaders` and substitutions, and the proxy enforces them.

A request outside a service's methods or paths gets a 403 with `blocked by service policy`, rather than being forwarded without its credential. The check runs on the request as it arrived, before anything rewrites it.

Path prefixes match whole segments and are compared against `EscapedPath()`, never a decoded one. A path-restricted service refuses outright anything that would have to be normalised to judge, which is what stops `/admin/%2e%2e/repos/x` reading as two different paths to us and to the upstream. A nil methods map means every method, so an unrestricted service is unaffected.

**Substitutions** run on the escaped path and write back escaped, and the replacement is escaped too, so a secret containing a slash cannot add a segment and a project addressed as `group%2Fproject` doesn't arrive as two segments. `RawQuery` had the same shape of bug: a base64 key containing `+` reached the upstream as a space.

The placeholder is matched in the escaped form as well as the form the author typed. `EscapedPath()` re-encodes the whole path whenever what the agent sent is not already valid encoding, so a `{{TOKEN}}` on the wire reads as `%7B%7BTOKEN%7D%7D`: matching only the literal left those placeholders on the wire, working in the query string and silently doing nothing in the path, with the agent seeing a third-party 404 and no warning.

**The post-substitution path re-check** no longer refuses the escaping this file itself writes. The replacement is `PathEscape`d precisely so a secret containing a slash cannot add a segment, and that `%2F` is what `isAmbiguousPath` reads as a separator, so a GitLab project addressed as `group%2Fproject` answered 403 against a prefix that plainly covered it, and the refusal blamed the path rule. The prefix comparison stays byte-exact, which is what holds the substituted span after the prefix, and the only thing still refused is traversal, judged on the decoded path because an upstream that decodes `%2F` before routing is the reader `..%2F..%2Fadmin` is written for.

**Injection order:** custom headers are written before the credential, so one naming the credential's own header loses to it instead of replacing the real token with a value the agent cannot explain. The backend refuses that pairing on write, so this is the floor under that check rather than a replacement for it: a service saved before the check existed, or a write that raced it, can no longer cost an agent its credential. Pass-through injects nothing, so a custom `Authorization` header on one still lands.

The resolve payload field is `customHeaders`, matching the server, so this and the server PR land together.

Server PR: https://github.com/Infisical/infisical/pull/8130

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


packages/agentvault/policy.go
```diff
@@ -0,0 +1,158 @@
...
+	return nil
+}
+
+// A nil map is every method. The set is built upper-case, and the comparison folds the request's method the
+// same way, so a client sending "get" is judged on GET rather than silently blocked.
+func (s *resolvedService) allowsMethod(method string) bool {
+	if s.allowedMethods == nil {
+		return true
+	}
+	return s.allowedMethods[strings.ToUpper(method)]
```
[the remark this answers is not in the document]
saifsmailbox98: False positive. The allowed set is built uppercase from the method enum, so `ToUpper(req.Method)` can only ever match a method the policy listed. `pOsT` or `delete` against a GET-only rule is still blocked, there is no casing that reaches a method you did not allow. Folding is deliberate so a client sending lowercase is not silently blocked instead.

packages/agentvault/rewrite.go
```diff
@@ -35,6 +48,181 @@ func injectCredential(req *http.Request, cred *credential) bool {
...
+	}
+
+	body, err := io.ReadAll(io.LimitReader(req.Body, maxBodyRewriteSize+1))
+	if err != nil {
+		// The stream is already part-consumed, so the original length is no longer true. Left alone,
+		// http.Transport refuses the request outright and the agent gets a 502 rather than the unchanged
+		// body this path promises.
+		_ = req.Body.Close()
+		req.Body = io.NopCloser(bytes.NewReader(body))
+		req.ContentLength = int64(len(body))
```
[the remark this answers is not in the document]
saifsmailbox98: Valid. Tested it: a body that breaks at 300 of 1007 bytes was reaching the upstream as a clean 300 byte request with a matching Content-Length, and the upstream returned 200. The declared length is left alone now, so the transport refuses and nothing is sent. Two lines, plus a test.

packages/agentvault/policy.go
```diff
@@ -0,0 +1,181 @@
...
+// and that escape is the '%2F' isAmbiguousPath refuses: judged by pathAllowed, a GitLab project addressed
+// as `group%2Fproject` would 403 against a prefix that plainly covers it.
+//
+// The prefix comparison is unchanged and still byte-exact, which is what keeps the substituted span after
+// the prefix: a placeholder sitting inside the prefix region rewrites those bytes and fails the comparison.
+// That leaves traversal as the only way out of an allowed prefix, so it is the only thing still refused,
+// and it is judged on the decoded path because an upstream that decodes '%2F' before routing is exactly
+// the reader `..%2F..%2Fadmin` is written for.
+func pathAllowedAfterSubstitution(escaped, decoded string, prefixes []string) bool {
+	for _, segment := range strings.Split(decoded, "/") {
```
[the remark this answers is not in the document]
saifsmailbox98: Valid, and it was a regression from loosening this re-check earlier in the PR. Confirmed both cases: with prefix `/repos`, a value starting with `\` or containing `;` walked out to `/admin` on IIS and Tomcat. Both are refused on the decoded path now, and `group%2Fproject` still passes. Test added.

packages/agentvault/rewrite.go
```diff
@@ -35,6 +48,196 @@ func injectCredential(req *http.Request, cred *credential) bool {
...
+	}
+	if enc := req.Header.Get("Content-Encoding"); enc != "" {
+		log.Warn().
+			Str("service", serviceName).
+			Str("contentEncoding", enc).
+			Msg("agent-vault: body substitution skipped on an encoded body; the placeholder is going upstream unchanged")
+		return false
+	}
+
+	body, err := io.ReadAll(io.LimitReader(req.Body, maxBodyRewriteSize+1))
```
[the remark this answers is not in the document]
saifsmailbox98: Valid. A declared length over the cap now skips the read entirely, so an oversized body costs no memory at all. The check after reading stays, since a chunked request declares -1 and a declared length is only a claim. Not adding a global budget for now, it needs a valid session token and the proxy it would kill is the agent's own.

saifsmailbox98:
@greptile please re-review.

Since the last pass:
- A placeholder is matched in its escaped form too, so `{{TOKEN}}` works in the path and not only in the query string.
- The post-substitution path re-check accepts the proxy's own `%2F` but now also refuses `;` and `\` on the decoded path, so a substituted value cannot walk out of its prefix on Tomcat or IIS.
- Custom headers are written before the credential, so one naming the credential's own header loses to it rather than replacing the real token.
- A body read that fails partway is no longer relabelled and forwarded; the declared length is left alone so the transport refuses it. A body whose declared length is already over the cap is never read at all.
- `headers` is `customHeaders` on the resolve payload, matching the server.

The method case-folding comment was answered inline as a false positive: the allowed set is uppercase, so no casing reaches a method the policy did not list.

Companion server PR: https://github.com/Infisical/infisical/pull/8130

packages/agentvault/rewrite.go
```diff
@@ -35,6 +48,205 @@ func injectCredential(req *http.Request, cred *credential) bool {
...
+			// encoding, so a `{{TOKEN}}` on the wire reads here as `%7B%7BTOKEN%7D%7D`. Looking only for the
+			// form the author typed would leave those placeholders on the wire, and the agent would see
+			// nothing but a third-party 404.
+			needle := sub.placeholder
+			if !strings.Contains(escaped, needle) {
+				needle = escapedPathForm(sub.placeholder)
+			}
+			if strings.Contains(escaped, needle) {
+				// The replacement is escaped too, or a secret containing '/' or '?' would itself reshape the URL.
+				if v, ok := replaceWithinLimit(escaped, needle, url.PathEscape(real), maxBodyRewriteSize); ok {
```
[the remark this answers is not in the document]
saifsmailbox98: Confirmed the example, `/items/a+b/a%2Bb` gives `/items/REAL/a%2Bb`. Not fixing it though. `a%2Bb` is neither the literal placeholder nor its escaped form, it is a third encoding the agent picked, and there are unlimited such forms. Covering them means decoding the path, substituting and re-encoding, which we deliberately do not do because it rewrites every other segment's byte form and breaks `group%2Fproject`. Normal cases are fine, a placeholder repeated in a path is replaced everywhere.

saifsmailbox98:
@greptile review

packages/agentvault/rewrite.go
```diff
@@ -35,6 +48,205 @@ func injectCredential(req *http.Request, cred *credential) bool {
...
+// The body is only ever read when some substitution names it, so a service without one keeps streaming
+// exactly as before.
+func applyBodySubstitutions(req *http.Request, serviceName string, subs []substitution) bool {
+	if req.Body == http.NoBody || req.ContentLength == 0 {
+		return false
+	}
+	if enc := req.Header.Get("Content-Encoding"); enc != "" {
+		log.Warn().
+			Str("service", serviceName).
+			Str("contentEncoding", enc).
```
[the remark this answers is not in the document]
saifsmailbox98: Valid. Header substitution runs before this, so `Content-Encoding: __TOKEN__` with a body would put the real secret in the proxy log, and the agent never had that secret to begin with. Logging only that an encoding was present now.

packages/agentvault/policy.go
```diff
@@ -0,0 +1,158 @@
...
+
+func (s *resolvedService) allowsMethod(method string) bool {
+	if s.allowedMethods == nil {
+		return true
+	}
+	return s.allowedMethods[strings.ToUpper(method)]
+}
+
+func requestPath(req *http.Request) string {
+	path := req.URL.EscapedPath()
```
scott-ray-wilson, line 104: Go’s EscapedPath() has a trap. If the path contains any byte Go thinks should have been encoded ({, }, |, ^, backtick, quote), it throws away what the agent actually sent and rebuilds the path from the decoded version, which turns %2F into a real /. hasUnsafeEscape exists specifically to refuse %2F. Adding one { anywhere in the path defeats it, because the %2F is already gone by the time the check runs.

packages/agentvault/rewrite.go
```diff
@@ -35,12 +47,194 @@ func injectCredential(req *http.Request, cred *credential) bool {
...
+		if len(sub.placeholder) == 0 {
+			continue
+		}
+		real := string(sub.value)
+
+		// Swapped in the escaped path so every other segment keeps the byte form the agent sent. Rewriting the
+		// decoded Path makes Go re-derive the wire path without re-escaping '/', and `group%2Fproject` would
+		// arrive as two segments pointing at a different resource.
+		if sub.surfaces[surfacePath] {
+			escaped := req.URL.EscapedPath()
```
scott-ray-wilson, line 92: Go’s EscapedPath() has a trap. If the path contains any byte Go thinks should have been encoded ({, }, |, ^, backtick, quote), it throws away what the agent actually sent and rebuilds the path from the decoded version, which turns %2F into a real /. hasUnsafeEscape exists specifically to refuse %2F. Adding one { anywhere in the path defeats it, because the %2F is already gone by the time the check runs.

packages/agentvault/proxy.go
scott-ray-wilson, line 431: The proxy swaps the placeholder for the real token on the way out, and nobody checks the response on the way back. The agent asks for /repos/{{PAT}}, the upstream answers with the near-universal “add a trailing slash” redirect containing the full URL, and this loop copies that header back verbatim. The agent now holds the raw credential. Grep for scrub/redact across the package returns nothing, so the absence of any scrubbing is certain; whether a given upstream echoes depends on the upstream.

scott-ray-wilson: Custom headers and substitutions sit inside the same https-only guard as the credential. A pass-through service whose only purpose is adding a header does nothing over http, and the single warning says the proxy refused to attach a credential, for a service that has no credential. Method and path rules run above that guard, so half the policy applies over http and half silently does not.

packages/agentvault/rewrite.go
```diff
@@ -35,12 +47,194 @@ func injectCredential(req *http.Request, cred *credential) bool {
...
+		req.Header.Set(header.name, value)
+	}
+	return len(customHeaders) > 0
+}
+
+// A body it cannot rewrite is logged rather than skipped in silence: the placeholder goes upstream and the
+// agent would otherwise see only a third-party 401.
+func applySubstitutions(req *http.Request, serviceName string, subs []substitution) []string {
+	changed := map[string]bool{}
+	for _, sub := range subs {
```
scott-ray-wilson, line 82: Each substitution runs on the output of the previous one, so if one placeholder is a prefix of another, the wrong token goes out with junk attached. Which one you get depends on the order the server returned the list in:
```
[__PAT__ first]   wire="/x/SHORT_LIVEDV2"
[__PAT__V2 first] wire="/x/FINE_GRAINED"
```

packages/agentvault/rewrite.go
```diff
@@ -35,12 +47,194 @@ func injectCredential(req *http.Request, cred *credential) bool {
...
+			Int64("declaredBytes", req.ContentLength).
+			Msg("agent-vault: body larger than the substitution limit; the placeholder is going upstream unchanged")
+		return false
+	}
+
+	body, err := io.ReadAll(io.LimitReader(req.Body, maxBodyRewriteSize+1))
+	if err != nil {
+		// ContentLength is deliberately left disagreeing with the bytes, so http.Transport refuses the request.
+		// Correcting it would hand the upstream a well-formed shorter request it cannot tell from a complete
+		// one, turning a broken upload into a partial write nobody can take back.
```
scott-ray-wilson: The safety net is “leave the byte count wrong so Go refuses to send it.” For a chunked upload the count is already -1, so there is nothing to leave wrong and Go sends it happily. Agent reproduction: the upstream received 27 of 1007 bytes with a real credential attached and answered 200, while the log said “refusing to forward a truncated one.”

packages/agentvault/rewrite.go
```diff
@@ -35,12 +47,194 @@ func injectCredential(req *http.Request, cred *credential) bool {
...
+}
+
+// Returns the input unchanged when the expansion would exceed limit, so a short placeholder mapped to a long
+// secret cannot balloon proxy memory.
+func replaceWithinLimit(s, old, replacement string, limit int) (string, bool) {
+	count := strings.Count(s, old)
+	if count == 0 {
+		return s, true
+	}
+	if len(s)+count*(len(replacement)-len(old)) > limit {
```
scott-ray-wilson: This multiplication overflows on 32-bit and wraps to a small or negative number, so the “is this too big?” check answers “no” for something enormous. Go then tries to allocate it and the whole process dies with fatal error: out of memory. .goreleaser.yaml ships 386 and armv6/armv7 binaries. Fix: compute in int64, or test count > (limit-len(s))/delta without multiplying.

packages/agentvault/rewrite.go
```diff
@@ -35,12 +47,194 @@ func injectCredential(req *http.Request, cred *credential) bool {
...
+	// Judged before reading, so an oversize body costs no memory. The check below still has to stand on its
+	// own: a chunked request declares -1, and a declared length is a claim rather than a fact.
+	if req.ContentLength > maxBodyRewriteSize {
+		log.Warn().Str("service", serviceName).Int("limitBytes", maxBodyRewriteSize).
+			Int64("declaredBytes", req.ContentLength).
+			Msg("agent-vault: body larger than the substitution limit; the placeholder is going upstream unchanged")
+		return false
+	}
+
+	body, err := io.ReadAll(io.LimitReader(req.Body, maxBodyRewriteSize+1))
```
scott-ray-wilson, line 243: The cheap pre-check above only looks at a declared Content-Length. A chunked request declares nothing, walks past it, and buffers 10 MB, with another copy per firing substitution. At 512 allowed connections that is several GB and an OOM kill, from one agent setting Transfer-Encoding: chunked.

packages/agentvault/proxy.go
scott-ray-wilson: This compares exactly, while the method allowlist 300 lines away deliberately upper-cases first. Send trace lowercase and Go passes it through untouched, so the credential gets injected into the one request type designed to echo your headers back. Also, when the guard does fire it returns before the logging call, making it the only security refusal in the package with no audit record.

packages/agentvault/policy.go
```diff
@@ -0,0 +1,158 @@
...
+		if i+2 >= len(escaped) {
+			return true
+		}
+		hi, hiOk := unhex(escaped[i+1])
+		lo, loOk := unhex(escaped[i+2])
+		if !hiOk || !loOk {
+			return true
+		}
+		b := hi<<4 | lo
+		if b < 0x20 || b == 0x7f {
```
scott-ray-wilson, line 217: The check is “less than 0x20” and a space is exactly 0x20, so it slips by. The segment check only rejects a segment that is exactly . or .., so ..%20 and .... read as ordinary names. On an upstream that strips trailing dots and spaces (Windows, IIS), /repos/..%20/admin resolves to /admin. .... needs no decoding at all.

packages/agentvault/policy.go
```diff
@@ -0,0 +1,158 @@
...
+	"net/http"
+	"strings"
+	"unicode/utf8"
+)
+
+var errPolicyBlocked = errors.New("blocked by service policy")
+
+// Runs before the plaintext refusal that nils a match, so a restriction holds on http:// too.
+func checkServicePolicy(svc *resolvedService, req *http.Request) error {
+	if !svc.allowsMethod(req.Method) {
```
scott-ray-wilson, line 19: Restrict a service to GET and POST, and the agent sends a POST carrying X-HTTP-Method-Override: DELETE. Nothing in the package strips that header (grep returns zero hits), so Rails, Symfony and Laravel all perform the DELETE. The path checks here are deliberately paranoid about upstream quirks; the method check applies none of the same reasoning.

packages/agentvault/rewrite.go
```diff
@@ -35,12 +47,194 @@ func injectCredential(req *http.Request, cred *credential) bool {
...
+						req.URL.RawPath = v
+						changed[surfacePath] = true
+					}
+				}
+			}
+		}
+
+		// Escaped, because RawQuery goes on the wire verbatim. A base64 key containing '+' would otherwise
+		// arrive as a space, and one containing '&' would split into a second parameter.
+		if sub.surfaces[surfaceQuery] && strings.Contains(req.URL.RawQuery, sub.placeholder) {
```
scott-ray-wilson: Every normal client sends {{TOKEN}} in a query as %7B%7BTOKEN%7D%7D. This line only looks for the literal typed form. url.Values.Encode() emits key=%7B%7BTOKEN%7D%7D; the result was surfaces=[] and the placeholder went to the third party untouched, unlogged. The typed form key={{TOKEN}} substitutes fine. The path branch 20 lines up handles this case; the query branch has no equivalent.

packages/agentvault/resolve.go
```diff
@@ -83,3 +88,67 @@ func toCredential(wire api.AgentVaultCredential) credential {
...
+		return nil
+	}
+	out := make([]string, 0, len(prefixes))
+	for _, prefix := range prefixes {
+		prefix = strings.TrimSpace(prefix)
+		if prefix == "" {
+			continue
+		}
+		if prefix != "/" {
+			prefix = strings.TrimRight(prefix, "/")
```
scott-ray-wilson, line 117: The “is it empty?” check runs four lines before the trim, so "//" survives the check, then trims down to "", and an empty prefix matches every path. A restricted service now allows everything, the opposite of the fail-closed promise in the comment above. The neighbours behave correctly, so this is specific to the all-slashes form:

```toPathPrefixes(["//"])  = [""]      pathAllowed("/admin/delete-org") = true
toPathPrefixes(["  "])  = ["\x00"]  pathAllowed("/admin/delete-org") = false
```

The server rejects // today, so this is defence-in-depth rather than a live hole, but the CLI is the enforcement point and re-validates nothing. One-line fix: trim before the empty check.

packages/agentvault/policy.go
```diff
@@ -0,0 +1,158 @@
...
+		return false
+	}
+	return matchesPrefix(escaped, prefixes)
+}
+
+// The path here is part-written by us: applySubstitutions escapes the value so it cannot add a segment, and
+// pathAllowed would refuse that very '%2F'. Only traversal can leave an allowed prefix, so only traversal is
+// refused, judged on the decoded path because that is what an upstream decoding '%2F' will route on. ';' and
+// '\' count as traversal here for the reason isAmbiguousPath gives.
+func pathAllowedAfterSubstitution(escaped, decoded string, prefixes []string) bool {
```
scott-ray-wilson, line 138: Two functions check the same path and disagree, and the weaker one guards the path that already carries the real secret. Wider than I first filed it; all four of these are refused before substitution and accepted after:

```
/repos//admin              pre=false  post=true
/repos/%00admin            pre=false  post=true
/repos/%09admin            pre=false  post=true
/repos/%C0%AE%C0%AE/admin  pre=false  post=true
```

An empty vault value turns /repos/__PAT__/admin into /repos//admin, and nothing validates that a substitution value is non-empty. The maintenance half matters too: lines 67 and 70-74 are hand-copies of 96 and 102-106, so hardening the first check silently skips the second.

packages/agentvault/policy.go
```diff
@@ -0,0 +1,158 @@
...
+		return true
+	}
+	return s.allowedMethods[strings.ToUpper(method)]
+}
+
+func requestPath(req *http.Request) string {
+	path := req.URL.EscapedPath()
+	if path == "" {
+		// OPTIONS * arrives as "*" and is left alone; a genuinely empty path is the root.
+		if req.URL.Opaque != "" {
```
scott-ray-wilson, line 108: http:admin/secrets through the tunnel gives Go an empty path and a non-empty Opaque. The plain-HTTP handler rejects that shape, but the tunnel handler calls forwardHTTP directly and nobody checks. With no prefixes configured (the default) it goes upstream as GET admin/secrets HTTP/1.1, no leading slash, with a real Authorization header. Also, the comment on line 37 about OPTIONS * is wrong: Go answers that itself and never calls the handler.

packages/agentvault/proxy.go
```diff
@@ -441,16 +445,29 @@ func (ps *proxyServer) blocksOffBundle(matched *resolvedService, hostname, port
...
+func (ps *proxyServer) forward(req *http.Request, scheme, hostname, port, sessionToken string) (*http.Response, *resolvedService, forwardOutcome, error) {
+	var outcome forwardOutcome
+
 	services, err := ps.cache.get(sessionToken)
 	if err != nil {
-		return nil, nil, fmt.Errorf("%w: %w", errSessionResolve, err)
+		return nil, nil, outcome, fmt.Errorf("%w: %w", errSessionResolve, err)
 	}
 
 	matched := bestMatch(services, hostname, port)
```
scott-ray-wilson, line 479: Service selection looks only at host and port, and on a policy failure it gives up instead of trying the next service. With api.github.com + /repos (read token) and api.github.com + /admin (admin token), whichever wins the host lookup handles every request and the other credential is unreachable behind a 403 that blames the path rule. packages/agentproxy/match.go:121 already threads the path into its matcher for this reason.

packages/agentvault/rewrite.go
```diff
@@ -35,12 +47,194 @@ func injectCredential(req *http.Request, cred *credential) bool {
...
+			if v, ok := replaceWithinLimit(req.URL.RawQuery, sub.placeholder, url.QueryEscape(real), maxBodyRewriteSize); ok {
```
scott-ray-wilson: QueryEscape is form encoding, which turns a space into +. Anything reading the raw query per RFC 3986 (AWS SigV4 signing, many gateways) sees a literal plus and the signature fails. The path surface next door gets it right:

```
query surface wire="/x?key=my+group"
path  surface wire="/x/my%20group"
```

The existing test cannot catch this because it asserts through url.ParseQuery, which converts + back to a space.

packages/agentvault/rewrite.go
```diff
@@ -35,12 +47,194 @@ func injectCredential(req *http.Request, cred *credential) bool {
...
+		real := string(sub.value)
+
+		// Swapped in the escaped path so every other segment keeps the byte form the agent sent. Rewriting the
+		// decoded Path makes Go re-derive the wire path without re-escaping '/', and `group%2Fproject` would
+		// arrive as two segments pointing at a different resource.
+		if sub.surfaces[surfacePath] {
+			escaped := req.URL.EscapedPath()
+			// EscapedPath re-encodes the whole path when what the agent sent is not already valid encoding, so
+			// `{{TOKEN}}` reads here as `%7B%7BTOKEN%7D%7D` and matching only the typed form would miss it.
+			needle := sub.placeholder
```
scott-ray-wilson, line 95: The fallback that handles %7B%7B builds its search string uppercase and compares byte for byte, so a client encoding in lowercase never matches and the substitution silently does not happen.
```
escapedPathForm("{{PAT}}") = "%7B%7BPAT%7D%7D"
/x/%7B%7BPAT%7D%7D -> "/x/SECRET"            surfaces=[path]
/x/%7b%7bPAT%7d%7d -> "/x/%7b%7bPAT%7d%7d"   surfaces=[]
```

scott-ray-wilson (changes requested):

packages/agentvault/resolve.go
saifsmailbox98: Valid, fixed by moving the empty check below the trim. You're right that it's defence-in-depth rather than live, since the server rejects // on write, but the CLI is the enforcement point so it should hold its own promise. Extended the existing "empty restriction allows nothing" test to cover // and /// alongside the whitespace case.

packages/agentvault/policy.go
saifsmailbox98: fixed. judged on the decoded segment rather than widening the byte rule — `..%20` decodes to `.. `, which is dots and spaces only, while `my%20repo` decodes to `my repo`, which is not. going from `< 0x20` to `<= 0x20` would have taken `/repos/my%20repo` with it, and that is already pinned as allowed.

`..%20`, `.. `, `....` and `.%20.` are all blocked now.

packages/agentvault/rewrite.go
saifsmailbox98: right, and the query fallback I pushed yesterday shipped with the same flaw, so both match case-insensitively now.

also made it say something when a substitution matches nothing. it was a miss rather than a leak — the placeholder goes upstream and the agent gets a 401 it cannot explain — but it was silent, which this file avoids everywhere else.

packages/agentvault/rewrite.go
saifsmailbox98: fixed, and you were right that the test could not have caught it. query values escape per rfc 3986 now so a space goes out as `%20`, and the test compares the wire query instead of going through `ParseQuery`.

left the needle on the line above alone since that matches the agent own encoding of the placeholder, and `{{...}}` has no space in it.

packages/agentvault/rewrite.go
saifsmailbox98: took the division form. confirmed the build targets too — goarch has 386 and arm with goarm 6 and 7, and the ignores only drop freebsd/386 and the windows ones.

reproduced at 32 bits: a 10MB body of a 5-char placeholder against a 1.2KB secret gives `len(s)+count*delta = -1778384896`, so the guard waves it through where int64 gives 2516582400 and refuses.

packages/agentvault/proxy.go
saifsmailbox98: both fixed. upper-cased like `allowsMethod` already was, so `trace` and `track` go the same way as the upper-case forms.

moved the refusal into `forward` so it goes through the same path as every other refusal and gets logged. side effect is it answers 403 rather than 405 now, which matches what the method allowlist refusal already returns — shout if you would rather keep 405 and log it separately.

packages/agentvault/policy.go
saifsmailbox98: fixed, the three override headers get stripped.

scoped it to services that restrict methods though, rather than always: with no restriction the agent can send DELETE outright, so the header gains it nothing and there is no control to protect. shout if you would rather it always strip.

packages/agentvault/proxy.go
saifsmailbox98: you're right, and I went further than moving the two out of the guard — the guard's gone entirely, so a credential attaches over http the same as over https.

the reason it existed doesn't really survive contact: refusing doesn't make anyone safer, it just breaks the feature and pushes the admin to stick the real credential in the agent's env instead, which is permanent and is the thing this product exists to stop. agent-proxy and the open source agent-vault both inject regardless of scheme, so we were the odd one out.

what keeps a credential off a plaintext wire now is the portless-means-443 rule — a bare host never matches port 80, so a service only brokers over http if someone typed the port. felt like the right place for it: the config then says on its face that the credential goes out unencrypted.

worth saying what it doesn't stop: an agent that opens a tunnel to a plaintext port and speaks cleartext inside gets the credential in the clear. deliberate, and both siblings do the same.

the log line's fixed too, and there was a second thing hiding behind it — nilling the match also dropped the service and bundle name from the request log, so a plaintext request to a configured service showed up as if nothing had matched. it's an info line with plaintext=true now rather than a warning, since everything else we warn about is a failure or a refusal, not a config working as written.

tested live against a real service with a bearer credential, a custom header and a path substitution over plain http. before: credential NONE, placeholder untouched, no header. after: all three land.

packages/agentvault/rewrite.go
saifsmailbox98: yeah, valid — fixed by sorting longest placeholder first before applying any of them, so `__PAT__V2` is always consumed before `__PAT__` can eat its prefix. order the server returns them in stops mattering.

reproduced your example both ways round first, it does exactly what you said.

packages/agentvault/rewrite.go
saifsmailbox98: valid, and fixed — the query branch matches the percent-encoded form now, same as the path branch already did.

one follow-up you'll see on another thread: the fallback I added here shipped with its needle built upper-cased and compared byte for byte, so `%7b%7bTOKEN%7d%7d` still missed. percent-escapes carry no required case and clients differ, so both surfaces fold escape case now. a substitution that matches nothing also says so in the log, which is what would have made this visible in the first place.

packages/agentvault/policy.go
saifsmailbox98: right, and worse than the `%2F` part alone — the path we forwarded wasn't the one the agent sent. tested it on the wire with a raw dial, service pinned to `/repos`:

```
/repos/a%2Fb              -> 403
/repos/a%2Fb/{x}          -> 200, upstream got /repos/a/b/%7Bx%7D
/repos/a%2F..%2Fadmin/{x} -> 403
```

so the agent asked for one segment `a%2Fb` and github got two, with the credential on it. traversal was still caught by the `..` and `//` checks on the rebuilt path, so it wasn't a prefix escape, but "we broker to a different resource than was asked for" is bad enough on its own.

couldn't just refuse a brace, since `{{PAT}}` is the placeholder syntax and that's the main path through this feature. so the bytes Go objects to get escaped before anything reads the path, which keeps RawPath valid and stops the rebuild happening at all.

two things worth knowing about the fix. it only runs where Go was going to rebuild anyway — `!`, `(`, `)`, `*`, `[`, `]` are rejected by shouldEscape but accepted by validEncoded, so escaping those would have changed the wire and broken a byte-compared prefix for no reason. and the predicate is derived from the stdlib rather than copied out of it, because the table behind encodePath is generated and a hand copy would drift a release later.

the wire does change in one place, deliberately: a percent-triple the agent wrote is now carried through instead of decoded, so `%41` survives the same way `%2F` does. that's the fix rather than a side effect.

same bug at rewrite.go, fixed by the same thing since it reads the corrected value.

packages/agentvault/rewrite.go
saifsmailbox98: same as the policy.go one — fixed by escaping the bytes Go objects to before anything reads the path, so this reads the corrected value and EscapedPath stops rebuilding. the comment here about EscapedPath re-encoding was describing the old cause, so that's rewritten too.

packages/agentvault/rewrite.go
saifsmailbox98: valid, and the reproduction matches. confirmed both halves:

```
declared 1007 -> body holds 28 bytes, count still says 1007 -> transport refuses
chunked  (-1) -> body holds 28 bytes, nothing to disagree with -> sent
```

so the trick only ever worked for a declared length. refusing outright now rather than relying on the byte count, which means the encoding stops mattering.

two details. only the read failure is an error — a skipped body, a Content-Encoding'd one and an oversized one stay the no-ops they were. and the path and headers are already rewritten by the time the body read fails, so the surfaces that fired come back alongside the error: the request is refused, but the log still has to say the credential was written into it. that second one I nearly got wrong.

it's the agent's own upload that broke, so it's a 400 with its own message rather than the generic 502 for an upstream we never reached.

packages/agentvault/policy.go
saifsmailbox98: you're right that nobody checks it, and the shape reaches forwardHTTP through the tunnel while the plain door turns it away — though by way of the empty host rather than any opaque check, so the two doors disagreed by accident rather than by design. refused in forwardHTTP now, where both of them meet.

one thing it isn't, though: an access bypass. with no prefixes configured the service allows every path anyway, so the agent could have sent `/admin/secrets` normally and got the same credential. host pinning also holds, so it can't reach anywhere else.

what it actually cost is the record. forwardHTTP logged `r.URL.EscapedPath()`, which is empty for an opaque target, so the policy check saw `admin/secrets`, the upstream saw `admin/secrets`, and the log line showed nothing. the one request shape that looks deliberate was the one you couldn't see afterwards. the logged path comes from requestPath now, which can't be blank.

and the `OPTIONS *` comment is wrong, as you said — checked the stdlib, `globalOptionsHandler` takes it before our handler runs, for both the front server and the per-tunnel one. it's misplaced as well as stale: `GET *` parses to Path `"*"`, which isn't empty, so the branch it annotates was never reachable for `*` either way. gone.

packages/agentvault/policy.go
saifsmailbox98: valid, and wider than the four you listed in one direction and narrower in another.

wider: the two functions disagree because the second one is weaker *on purpose*, and that wasn't written down. we PathEscape the substituted value so it can't add a segment, and isAmbiguousPath refuses that very escaping — I tried unifying them and it 403s our own output:

```
secret "org/repo" -> /repos/org%2Frepo/admin -> refused
secret "a%b"      -> /repos/a%25b/admin      -> refused
```

so they have to stay separate. what the second one was missing is the shapes a *value* can introduce, judged on the decoded path where our own escaping is invisible. `//`, control bytes and invalid UTF-8 are refused there now, and `ghp_plain`, `org/repo`, `v1.2`, `a%b` all still pass.

narrower: your empty-value trigger doesn't fire. `secretValueSchema` is `.min(1)` and blocks control chars, and the agent's half of the path has already been through isAmbiguousPath on arrival, so none of the four shapes can actually be produced today. it was latent, held shut by validation in the other repo, which isn't something this package should be leaning on — hence fixing it anyway.

one behaviour change worth flagging: a secret with a leading or trailing slash now fails this check rather than reaching the upstream as a doubled separator. `/admin` in `/repos/{{PAT}}/x` used to go out as `/repos//admin/x`. tested.

and yes on the maintenance half — the comment now says why the two differ, so it doesn't read as a hand-copy that drifted.

packages/agentvault/rewrite.go
saifsmailbox98: pushing back on this one, gently.

the cap is the same for both encodings. a chunked body walks past the declared-length pre-check, but it then hits `io.ReadAll(io.LimitReader(req.Body, maxBodyRewriteSize+1))` and is bounded at 10MB like everything else. so 512 × 10MB is equally reachable with declared lengths — send 512 requests each declaring 9.9MB and you get the same number.

what chunked actually loses is the *cheap* rejection: for a declared oversize we skip the read entirely and pay nothing, and for chunked we always read up to the cap before bailing. that's a real difference, just not one that raises the ceiling.

so if several GB in flight is the worry, the fix is a global in-flight budget across concurrent requests rather than a chunked special case — that's its own piece of work and I'd rather not squeeze it in here. happy to be told otherwise if you think the asymmetry alone is worth closing.

the truncation half of this, on the other thread, was real and is fixed.

packages/agentvault/proxy.go
saifsmailbox98: I don't think this one can happen, at least not yet.

two services on one host in one bundle is a hard reject at write time — tried building your exact example on a dev stack:

```
POST .../services  name=gh-read  api.github.com  /repos   -> created
POST .../services  name=gh-admin api.github.com  /admin   -> 400
  "'gh-read' already covers api.github.com:443 in this access bundle."
```

the conflict rule is host-only on purpose: methods and prefixes are filters on a service that already matched, not part of the match key. across bundles it's allowed, but `AGENT_VAULT_MAX_SESSION_BUNDLES` is 1, so a session carries one bundle and the two never meet.

the agentproxy citation is right but the reason differs — it allows overlapping services, so it has to disambiguate at request time. we forbid the overlap instead, and that's what makes host+port selection safe here.

it does become real the day multi-bundle sessions ship, since two bundles could each cover the same host and first-match would quietly pick one. worth remembering as a prerequisite for that rather than a fix now.

packages/agentvault/proxy.go
saifsmailbox98: reproduced it, and it's as bad as you'd expect — an ordinary 301 is enough, no hostile upstream needed. substitution value `ghp_SUPERSECRET`:

```
Location:       http://.../repos/ghp_SUPERSECRET/
X-Request-Path: /repos/ghp_SUPERSECRET
body:           {"error":"not found: /repos/ghp_SUPERSECRET"}
```

not changing it here though. a substitution into a path or a query puts the secret into a URL, and that's the admin's call at configuration time — the upstream logs it, proxies on the path see it, and getting it back in a redirect is the same exposure they already took on. we modify the request and we don't touch the response, and I'd rather that stay one simple rule than become "we scrub some of the ways it can come back".

worth knowing neither agentproxy nor the open-source agent-vault scrubs either, so it isn't a regression against the siblings.

if we do want to close it later, header scrubbing catches the redirect case cheaply and the body needs a rolling scan over the stream, gated on the request having actually substituted. not for this PR.

packages/agentvault/proxy.go
```diff
@@ -247,7 +247,8 @@ func (ps *proxyServer) handleConnect(w http.ResponseWriter, r *http.Request) {
 	// Node's fetch tunnels http:// targets too and speaks plaintext inside, where every other client sends
 	// absolute-form. A tunnel is opaque bytes to an ordinary proxy, so that works everywhere else; here the
 	// first byte decides: a TLS record starts with 0x16, anything else is plain HTTP and takes the same path
-	// as absolute-form, which already refuses to attach a credential over plaintext.
+	// as absolute-form. The port stays the one the CONNECT line named, so a service still only matches if
+	// its pattern covers that port.
```
scott-ray-wilson, line 251: The port opt-in doesn't hold on this path. An agent that CONNECTs to a service's ordinary port and then speaks plaintext instead of handshaking gets the credential attached in the clear, with no plaintext port ever configured.

Just below, anything whose first byte isn't `0x16` goes to `serveTunnel` with scheme `"http"` and the port from the CONNECT line. For any ordinary target that port is 443, which is exactly what a service that never named a port defaults to via `defaultPort`. So `bestMatch` matches on 443, and with the https guard gone `forward` injects into a plaintext request.

Drove it end to end through the real proxy. `hostPatterns` hand-built as `{host, port, portWritten: false}`, which is the shape a service that never named a port gets, with the fixture's port standing in for 443 so the forward could actually land. Nothing else on the path differs:

```
CONNECT -> 200 Connection Established
{"decision":"brokered","status":200,"service":"github","plaintext":true,...}
UPSTREAM SAW: path="/repos/x" Authorization=[Bearer tok_real]
```

In production that is `CONNECT api.github.com:443`, then a plaintext `GET` inside the tunnel, and the real token goes onto a TCP socket in cleartext. A TLS-only upstream won't answer it, but the bytes have already left the machine.

The `plaintext: true` log field is a good addition and means this is at least auditable. What's missing is the enforcement. `proxy_plaintext_test.go` states the invariant as "A service reaches this fixture only by naming its port, which is what opts it into plaintext", and all three of its cases drive the absolute-form leg, so nothing pins the tunnel.

Two small ways to close it: refuse a non-TLS first byte inside a tunnel outright, or gate plaintext brokering on `portWritten` rather than on port equality, so a service sitting on the 443 default can't be reached this way.

packages/agentvault/rewrite.go
```diff
@@ -35,12 +47,270 @@ func injectCredential(req *http.Request, cred *credential) bool {
...
+	replaced := false
+	for _, sub := range subs {
+		if !sub.surfaces[surfaceBody] || len(sub.placeholder) == 0 {
+			continue
+		}
+		count := bytes.Count(rewritten, []byte(sub.placeholder))
+		if count == 0 {
+			continue
+		}
+		if len(rewritten)+count*(len(sub.value)-len(sub.placeholder)) > maxBodyRewriteSize {
```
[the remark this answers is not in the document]
saifsmailbox98: fixed. same multiplication `replaceWithinLimit` was moved off, just in the body path, so it divides now too.

worth knowing it's reachable with an ordinary config rather than a hostile one: the placeholder minimum is 1 byte on purpose and a value can be 8KB, so a body full of that placeholder is all it takes. what dies is the agent's own proxy, which is why this is three lines rather than a global budget.

the test that comes with it can only pin the refusal, not the overflow — on a 64-bit runner the old form refuses the body correctly too. checked the 386 and armv6 builds still compile.

packages/agentvault/proxy.go
```diff
@@ -463,24 +498,51 @@ func (ps *proxyServer) forward(req *http.Request, scheme, hostname, port, sessio
...
-		} else {
-			injectCredential(req, &matched.credential)
+		// Substitutions first, so an injected real value can never itself be rewritten. The credential last,
+		// so a custom header naming the credential's own header loses rather than replacing the token.
+		substituted, err := applySubstitutions(req, matched.name, matched.substitutions)
+		outcome.substituted = substituted
+		if err != nil {
+			return nil, matched, outcome, err
+		}
+		outcome.brokered = injectCustomHeaders(req, matched.customHeaders)
```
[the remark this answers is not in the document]
saifsmailbox98: fixed, the strip runs after the brokered headers now rather than before them, so it no longer matters who set the header.

one thing the report has backwards: the session holder can't reach this. their own override header was already being stripped. it needed an admin to configure `X-HTTP-Method-Override` as a custom header on a service that also restricts methods, which is a config contradicting itself. still worth closing since the guarantee shouldn't depend on ordering.

test added, it fails without the move.

packages/agentvault/policy.go
```diff
@@ -0,0 +1,203 @@
...
+	if s.allowedMethods == nil {
+		return true
+	}
+	return s.allowedMethods[strings.ToUpper(method)]
+}
+
+// Rails, Laravel and Symfony all honour these, so a POST carrying one performs the method it names. The
+// wire method is what the allowlist judged, so where there is an allowlist the header has to go.
+func stripMethodOverrideHeaders(header http.Header) {
+	for _, name := range []string{"X-HTTP-Method-Override", "X-Method-Override", "X-HTTP-Method"} {
```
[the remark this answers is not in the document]
saifsmailbox98: not doing this one.

`_method` isn't HTTP, it's a Rails/Laravel/Symfony convention for getting past what old HTML forms could send. the request we forward is an honest POST; the upstream decides to read a field in the body and act as if it were something else. by that logic `?action=delete` or a json `{"op":"delete"}` would need chasing too, and a GraphQL API takes every mutation over POST, so a GET/POST-only rule never constrained deletes there either.

so a method rule means the method on the wire, not whatever the upstream decides that method means. the override headers are stripped because they were asked for by name in review, not because the line is drawn there.

packages/agentvault/proxy.go
```diff
@@ -463,24 +485,47 @@ func (ps *proxyServer) forward(req *http.Request, scheme, hostname, port, sessio
...
-			log.Warn().
-				Str("host", hostname).
-				Str("service", matched.name).
-				Msg("agent-vault: refusing to attach a credential over plaintext http")
-			matched = nil
-		} else {
-			injectCredential(req, &matched.credential)
+		// Substitutions first, so an injected real value can never itself be rewritten. The credential last,
+		// so a custom header naming the credential's own header loses rather than replacing the token.
+		outcome.substituted = applySubstitutions(req, matched.name, matched.substitutions)
```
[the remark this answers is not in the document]
saifsmailbox98: deliberate, and it's a change we made during this review rather than something that slipped.

the old guard refused everything over http, not just credentials: a pass-through service whose whole job is adding a header did nothing at all, and the log announced a refusal to attach a credential that service never had. both sibling products inject regardless of scheme.

a service only reaches plaintext if its host pattern names a plaintext port, which is the admin opting in — `match.go` carries that reasoning. the default stays 443.

packages/agentvault/proxy.go
saifsmailbox98: reproduced exactly, and you're right it was a real leak — the log even showed the credential substituted into the path before the dial, so on a live host it went out in the clear on a service that named no port.

took your first suggestion, the injection-time gate, rather than the tunnel sniff: `forward` now refuses plain HTTP to port 443 before substitution runs. it covers the tunnel path and the `http://host:443` absolute-form typo in one place, and can't be reopened by a future path that manufactures http-on-443. scoped to the literal 443 default rather than `portWritten`, so a service that names a plaintext port is untouched. both cases are pinned now — the two tests attach the credential without the gate.

here's the whole space so it's clear what's covered and what's deliberate. "exposed?" is whether anyone on the proxy→upstream wire can read the token:

| # | service config | agent request | proxy → upstream | exposed? | result |
|---|---|---|---|---|---|
| 1 | `host` (defaults 443) | CONNECT 443 + TLS | HTTPS | no | normal case |
| 2 | `host` (defaults 443) | CONNECT 443, then plain HTTP | **403** | no | **fixed** — this was the leak |
| 3 | `host` (defaults 443) | `http://host:443/...` | **403** | no | **fixed** — same path, typo'd URL |
| 4 | `host` (defaults 443) | `http://host/...` (port 80) | no match, no credential | nothing to expose | safe |
| 5 | `host:80` | plain HTTP | HTTP + credential | **yes** | by design — naming a port is the opt-in, documented |
| 6 | `host:80` | TLS | HTTPS, upstream can't complete | no | fails safe |
| 7 | `host:9000` | TLS | HTTPS | no | works |
| 8 | `host:9000` | plain HTTP | HTTP + credential | **yes** | by design — same opt-in |

what's deliberately not solved: rows 5 and 8. a written port is the admin saying "this service is plaintext, I know", so we can't both offer plaintext brokering and refuse to broker over plaintext. row 8 is the sharper one — for a nonstandard port only the agent knows whether the upstream speaks TLS, so we follow its request; the admin accepted that the moment they typed the port. the docs already warn the credential crosses that hop unencrypted.

one thing I didn't do: silently upgrade rows 2-3 to TLS instead of refusing. the only clients that produce them are broken or probing, and without a proxy they fail at the upstream anyway, so a 403 that names the reason keeps a broken agent from looking like it works until the day it runs without us.

packages/agentvault/rewrite.go
```diff
@@ -35,12 +47,273 @@ func injectCredential(req *http.Request, cred *credential) bool {
...
+			}
+		}
+
+		if sub.surfaces[surfaceHeader] {
+			for name, values := range req.Header {
+				for i, v := range values {
+					if !strings.Contains(v, sub.placeholder) {
+						continue
+					}
+					if replaced, ok := replaceWithinLimit(v, sub.placeholder, real, maxBodyRewriteSize); ok {
```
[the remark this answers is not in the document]
saifsmailbox98: this is a real property, and it's the deliberate shape of the product rather than something specific to substitutions. we swap one way only, placeholder to real value on the request out; responses pass through untouched. containment is the service's own methods and paths — an agent restricted to the endpoints it needs can't freely aim a placeholder at an echo endpoint.

worth noting the same is true of the credential itself, which is the more basic case: any service that reaches a header-echoing endpoint reveals its token to the agent, substitution or not. so "bind each substitution to a fixed location" wouldn't close the underlying thing, and it would break the documented usage where the agent places the placeholder where its request needs it.

so: not changing it, same reasoning as the equivalent thread on the backend PR.

packages/agentvault/rewrite.go
```diff
@@ -35,12 +47,284 @@ func injectCredential(req *http.Request, cred *credential) bool {
...
+		// secret reused across several headers. Scoped to the value the admin set, never the agent's request
+		// or the credential, so nothing the agent sends can steer it.
+		for _, sub := range subs {
+			if len(sub.placeholder) == 0 {
+				continue
+			}
+			if replaced, ok := replaceWithinLimit(value, sub.placeholder, string(sub.value), maxBodyRewriteSize); ok {
+				value = replaced
+			}
+		}
```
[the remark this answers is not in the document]
saifsmailbox98: fixed — good catch, and it's the one I was uneasy about. gated on the header surface now: a substitution resolves inside a custom header value only if the admin ticked headers, which is them saying the secret may appear in a header. a custom header is a header, so the same tick covers both it and the agent's own headers, and a body-only or path-only substitution leaves a colliding header value alone. two tests, the positive and the negative.

went with the surface tick rather than your literal suggestion because `surfaceHeader` already means exactly "this secret may go in a header", so it reads as one intent rather than two.

packages/agentvault/proxy.go
```diff
@@ -463,24 +498,61 @@ func (ps *proxyServer) forward(req *http.Request, scheme, hostname, port, sessio
 	// Stripped before injecting, so a client's Connection header cannot delete the credential.
 	stripHopByHopHeaders(req.Header)
 
+	// A service reaches port 443 by default, which everywhere else in the product means TLS, so plain HTTP
+	// here is either a tunnel that declined to handshake or an http:// URL naming 443 — nothing legitimate.
+	// Refused before substitution, which would otherwise put a secret in the path of a cleartext request.
+	if matched != nil && scheme == "http" && port == "443" {
+		return nil, matched, outcome, fmt.Errorf(
+			"service %q expects TLS on port 443; refusing to broker plain HTTP: %w", matched.name, errPolicyBlocked)
+	}
```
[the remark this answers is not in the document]
saifsmailbox98: deliberate, and the same call as the veria and human threads on this PR. the default 443 case is guarded — plain HTTP there is refused. a non-443 port only carries a credential because an admin wrote that port, which is the opt-in: it's how you point a service at an API that terminates TLS at the edge or serves no HTTPS at all. keeping injection HTTPS-only would delete that feature, which is the thing another reviewer asked for in the first place.

for a nonstandard port only the agent knows whether the upstream even speaks TLS, so we follow its request; the admin accepted the cleartext hop by writing the port, and the docs say so.

scott-ray-wilson (approved):
lgtm

PR #403, opened 2026-09-17
feat: pam built-in masking

sheensantoscapadngan:
# Description 📣
This PR adds support for built-in PAM masking
image

image

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


packages/pam/session/masking/detection.go
```diff
@@ -0,0 +1,94 @@
...
+	})
+
+	// Columns describe the match, not the secret, and diverge when a rule sets secretGroup, so
+	// there are no usable offsets to cut on.
+	result := s
+	for _, finding := range findings {
+		if len(finding.Secret) < minSecretLength {
+			continue
+		}
+		result = strings.ReplaceAll(result, finding.Secret, Placeholder)
```
lb-vn, line 79: The logger masks one rendered terminal line at a time, so anything spanning many lines never matches. Confirmed `cat id_rsa` still leaks the key body with detection on, since the private-key rule needs the whole BEGIN/END block in one fragment. credentialValues has the same gap for c.PrivateKey.

packages/pam/session/masking/credentials.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+		seen[v] = struct{}{}
+		secrets = append(secrets, v)
+	}
+	if len(secrets) == 0 {
+		return nil
+	}
+	// Longest first: a credential containing another (a passphrase inside a key block) must be
+	// redacted before the shorter one rewrites the text around it.
+	for i := 1; i < len(secrets); i++ {
```
lb-vn, line 55: A bit of a nit but Claude says:
we should just use sort.SliceStable here

packages/gateway-v2/capabilities.go
```diff
@@ -0,0 +1,5 @@
+package gatewayv2
+
+// Declared on each heartbeat. The platform stores the map verbatim, so an absent key means the
+// gateway is too old to support the feature, not that the feature is off.
```
lb-vn, line 4: A bit of a nit but I think this comment is useless

packages/pam/session/masking/masking_test.go
```diff
@@ -0,0 +1,252 @@
+package masking
```
lb-vn, line 1: If we end up fixing the multi-line detection for stuff like private keys, we should add a test case for that here.

lb-vn (changes requested):

lb-vn (approved):

